Issue #33, opened 2020-12-13
Find workaround for github api limitation of 1000 repos in results, expand sample set beyond 1000 by stars before sorting by criticality score


inferno-chromium:

tgamblin:
Is GHTorrent (developed by @gousiosg for software analytics) a potential source of data for the criticality score?  See https://ghtorrent.org/.  There are instructions for access to the database here: https://ghtorrent.org/raw.html.

gousiosg:
Hi @tgamblin thanks for the plug :-) GHTorrent is not suitable for this as it may have data freshness and general quality issues.

mfridman:
Would external sites (with an API) for specific languages help? I.e., I built a website that aggregates GitHub data for the top Go repositories with >50 stars. Its ~15k repos and tracks ~9mil stars (both increase/decrease)

The freshness of the data is 1 hour.

I'd imagine other websites such as bestofjs.org would have "fresh" data.

Depending on the threshold set, you can get quite a bit of repos from the search API .. is this currently the biggest hurdle?

---

I've ported this Python project to Go but it turns out in its current implementation the number of hits against the API limits is quite high per single repo, esp. the open/closed issue count. So I've now switched to optimizing it with a combination of REST V3 and GraphQL V4 (which iirc count as distinct rate limits).

My intention is to add a criticality score to the 15k Go repos tracked by said website.

inferno-chromium:
> Depending on the threshold set, you can get quite a bit of repos from the GitHub search API .. is this currently the biggest hurdle?

I am thinking on workarounds, maybe by partitioning sets by stars count, etc. How are you doing this on your website.

> 
> Would external sites (with an API) for specific languages help? I.e., I built a website that aggregates GitHub data for the top Go repositories with >50 stars. Its ~15k repos and tracks ~9mil stars (both increase/decrease)
> 
> The freshness of the data is 1 hour.

Sure, can you expose an api end point that return a github repo list (e.g. kubernetes/kubernetes) when given a language by stars.

> 
> I'd imagine other websites such as bestofjs.org would have "fresh" data.
> 
> I've ported this Python project to Go but it turns out in its current implementation the number of hits against the API limits is quite high per single repo, esp. the open/closed issue count. So I've now switched to optimizing it with a combination of REST V3 and GraphQL V4 (which iirc count as distinct rate limits).
> 
> My intention is to add a criticality score to the 15k Go repos tracked by said website.

[...]

PR #38, opened 2020-12-14
Workaround for using arbitrary sample size, beyond 1000 github results.

inferno-chromium:
Fixes https://github.com/ossf/criticality_score/issues/33

### DIFF
--- a/criticality_score/generate.py
+++ b/criticality_score/generate.py
@@ -61,7 +61,7 @@ def main():
             s = 1
             last_stars_processed = None
             while s <= args.sample_size:
-                query = f'language:{github_lang}'
+                query = f'language:{github_lang} archived:false'
                 if last_stars_processed:
                     query += f' stars:<{last_stars_processed+1}'
                 print(f'Running query: {query}')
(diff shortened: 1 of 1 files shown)


criticality_score/generate.py
```diff
@@ -48,34 +48,40 @@ def main():
     parser.add_argument(
         "--sample-size",
         type=int,
-        default=1000,
+        default=5000,
```
tgamblin, line 51: Just curious -- is there some rationale for this number (or the old number)? Any star threshold is going to mean that stars can potentially affect the criticality score, which seems a little sad since criticality score is supposed to go beyond stars.

Obviously, they're not independent, though. It seems like the default sample size should be set based on the count, though exactly how might vary by language. A thought: plot something like "likelihood that the top projects ranked by stars are in the top projects by criticality score" (based on the data you have so far), fit that curve, and set the sample size based on where the likelihood approaches zero.

inferno-chromium: Yes it is sad, but we are constantly scaling up, trying to process more repos so that we can do this better. github api is slow and has constant rate limits, so even a few thousand takes more a day.

It does not impact criticality score algo, which was the main purpose for project launch. We want organizations to generate criticality scores for their own oss usage, with their own infra.

PR #51, opened 2020-12-19
add Numerical Scientific software languages

Nelson-numerical-software:


criticality_score/generate.py
```diff
@@ -28,10 +28,16 @@
     'go': ['Go'],
     'java': ['Java', 'Groovy'],
     'js': ['Javascript', 'Typescript', 'CoffeeScript'],
+    'julia': ['Julia'],
+    'matlab': ['Matlab'],
+    'nelson': ['Nelson'],
```
inferno-chromium, line 33: i dont see Nelson, Octave, OpenMatrix in github languages.
Also, Matlab, scilab, julia, does it make sense to club in one category like scientific, computation or something ?

inferno-chromium (changes requested):

PR #57, opened 2020-12-26
Rename chash to csharp

coni2k:
Tiny fix for C# language:
https://en.wikipedia.org/wiki/C_Sharp_(programming_language)

### DIFF
--- a/criticality_score/generate.py
+++ b/criticality_score/generate.py
@@ -135,7 +135,7 @@ def main():
         stats.append(output)
 
     languages = '_'.join(args.language)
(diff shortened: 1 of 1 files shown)


inferno-chromium:
Csharp Results should be available in a few days.

coni2k:
I'm running the query at the moment, if it's any help?

inferno-chromium:
> I'm running the query at the moment, if it's any help?

Available at gs://ossf-criticality-score/csharp_top_200.csv

PR #60, opened 2021-01-02
Make language parameter optional

coni2k:
I wanted to try whether I could make language parameter optional, so it can just list all the repos.

I didn't try it with a large sample size but it looks like it's doing the job now.
So, I wanted to share it but feel free to discard it (or make necessary changes) of course.

### DIFF
--- a/criticality_score/generate.py
+++ b/criticality_score/generate.py
@@ -36,7 +36,19 @@ LANGUAGE_SEARCH_MAP = {
 }
 IGNORED_KEYWORDS = ['book', 'course', 'docs', 'interview', 'tutorial']
 
-def get_repo_urls(urls, sample_size, github_lang = ''):
+def get_github_repo_urls(sample_size, languages):
+    urls = []
+    if (languages):
+        for lang in languages:
+            lang = lang.lower()
+            for github_lang in LANGUAGE_SEARCH_MAP.get(lang, lang):
+                urls = get_github_repo_urls_for_language(urls, sample_size, github_lang)
+    else:
+        urls = get_github_repo_urls_for_language(urls, sample_size)
+
+    return urls
+
+def get_github_repo_urls_for_language(urls, sample_size, github_lang=None):
     """Return repository urls given a language list and sample size."""
     samples_processed = 1
     last_stars_processed = None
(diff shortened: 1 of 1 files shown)


criticality_score/generate.py
```diff
@@ -111,7 +110,16 @@ def main():
     repo_urls = set()
     for rnd in range(1, 4):
         print(f'Finding repos (round {rnd}):')
-        repo_urls.update(get_repo_urls(args.language, args.sample_size))
+        urls = []
+        if (args.language):
```
inferno-chromium, line 114: Can you move lines 114-120 into a helper function get_github_repo_urls(languages), will be adding gitlab in future, so moving this into a helper is better.

criticality_score/generate.py
```diff
@@ -36,48 +36,47 @@
 }
 IGNORED_KEYWORDS = ['book', 'course', 'docs', 'interview', 'tutorial']
 
-
-def get_repo_urls(languages, sample_size):
+def get_repo_urls(urls, sample_size, github_lang = ''):
```
inferno-chromium, line 39: s/get_repo_urls/get_github_repo_urls_for_language

and github_lang=None ('' is not needed).

inferno-chromium (changes requested):
Thanks, some minor changes needed to make future refactoring with gitlab repo easier.

criticality_score/generate.py
coni2k: Okay, it's now like this:

```
repo_urls.update(get_github_repo_urls(args.sample_size, args.language))

def get_github_repo_urls(sample_size, languages):
    urls = []
    if (languages):
        for lang in languages:
            lang = lang.lower()
            for github_lang in LANGUAGE_SEARCH_MAP.get(lang, lang):
                urls = get_github_repo_urls_for_language(urls, sample_size, github_lang)
    else:
        urls = get_github_repo_urls_for_language(urls, sample_size)

    return urls
```

criticality_score/generate.py
coni2k: This one is also updated

```
def get_github_repo_urls_for_language(urls, sample_size, github_lang=None):
```

inferno-chromium (approved):

PR #61, opened 2021-01-03
output folder update

coni2k:
In Readme, we don't mention required "output" parameter in the example, so I added that one.
Also created an empty output folder with a gitignore file in it. This one is optional of course; I can remove or rename the folder.


output/.gitignore
```diff
@@ -0,0 +1,4 @@
+# Ignore everything in this directory
```
inferno-chromium, line 1: Instead of adding this file, add this directory in the root directory's .gitignore.

inferno-chromium (changes requested):

output/.gitignore
coni2k: git doesn't create empty folders when we initially clone the repo. So, this is one of the work-arounds:
https://stackoverflow.com/questions/115983/how-can-i-add-an-empty-directory-to-a-git-repository

Actually, as the next step, what do you think of removing "output-dir" parameter from the script? So it always creates the csv files in "output" folder.

The problem is that it tries to create the files at the end of the script and if the "output-dir" is not there, it throws an exception and the whole execution fails:

```shell
Traceback (most recent call last):
  File "C:\Users\Serkan\AppData\Local\Programs\Python\Python36-32\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "C:\Users\Serkan\AppData\Local\Programs\Python\Python36-32\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "D:\Development\Projects\GitHub\Temp\criticality_score\criticality_score\generate.py", line 163, in <module>
    main()
  File "D:\Development\Projects\GitHub\Temp\criticality_score\criticality_score\generate.py", line 152, in main
    with open(output_filename, 'w') as file_handle:
FileNotFoundError: [Errno 2] No such file or directory: 'output\\all_top_1.csv'
```

inferno-chromium (approved):

output/.gitignore
inferno-chromium: Your patch is fine then, merging.

PR #63, opened 2021-01-04
Add stars column

coni2k:
When I was looking at the results, I noticed that it would be useful to have "stars" info as well.

The update works with GitHub, but I got an exception with GitLab due to datetime fields (will report it separately). So, couldn't make a final test with that one.

Again, up to you to accept it of course 🙏


inferno-chromium (commented):
Don't like stars at all, also putting it makes it confusing that we use in criticality score when we don't

coni2k:
Sure, no problem! Will close it then 👍

PR #66, opened 2021-01-08
Add logging to generate script

coni2k:
When we run generate script for large number of repos, it becomes difficult to follow the exceptions.
Not sure what are the best practices but I added basic logging by following this example. Looks like it's doing the job:
https://stackoverflow.com/questions/15474095/writing-a-log-file-from-python-program

As usual, feel free to accept/discard it

### DIFF
--- a/criticality_score/generate.py
+++ b/criticality_score/generate.py
@@ -15,12 +15,14 @@
 
 import argparse
 import csv
+import logging
 import os
 import sys
 import time
-import logging
 
 from . import run
+
+logger = logging.getLogger()
 
 LANGUAGE_SEARCH_MAP = {
     'c': ['C'],
@@ -62,7 +64,7 @@ def get_github_repo_urls_for_language(urls, sample_size, github_lang=None):
         if last_stars_processed:
             # +100 to avoid any races with star updates.
             query += f' stars:<{last_stars_processed+100}'
-        print(f'Running query: {query}')
+        logger.info(f'Running query: {query}')
         token_obj = run.get_github_auth_token()
         new_result = False
         repo = None
@@ -80,7 +82,7 @@ def get_github_repo_urls_for_language(urls, sample_size, github_lang=None):
                 continue
             urls.append(repo_url)
             new_result = True
-            print(f'Found repository'
+            logger.info(f'Found repository'
                     f'({samples_processed}): {repo_url}')
             samples_processed += 1
             if samples_processed > sample_size:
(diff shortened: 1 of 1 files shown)


criticality_score/generate.py
```diff
@@ -18,6 +18,7 @@
 import os
 import sys
 import time
+import logging
```
inferno-chromium, line 21: alpha order, move to before "import os"

criticality_score/generate.py
```diff
@@ -135,8 +147,9 @@ def main():
                 output = run.get_repository_stats(repo)
                 break
             except Exception as exp:
-                print(
-                    f'Exception occurred when reading repo: {repo_url}\n{exp}')
+                msg = f'Exception occurred when reading repo: {repo_url}\n{exp}'
+                print(msg)
```
inferno-chromium, line 151: This is not the right use of logging methods, to just log errors and not anything else. I think you can add logging handler, but then convert all the prints to use that logging pipelines (regular print should be info level). Then log file will have all log data, info and errors. and you can search for errors using error log level.

inferno-chromium (changes requested):

criticality_score/generate.py
coni2k: I see, since logger also prints the messages to the console.

Updates in last commit:
- Move logger to global
- Replace print with logger.info
- Set loglevel to info

I also tried to share the logger with `run` script. The only way that I could do it was to pass it as an argument in each function. e.g. `def get_repository(url, logger):`

I assume that there's a better way to do it, so I'm skipping that one for now.

criticality_score/generate.py
```diff
@@ -92,6 +95,15 @@ def get_github_repo_urls_for_language(urls, sample_size, github_lang=None):
 
 
 def main():
+    # logging
```
inferno-chromium, line 98: Please move all logging code to a helper function, initialize_logging_handlers

criticality_score/generate.py
```diff
@@ -92,6 +95,15 @@ def get_github_repo_urls_for_language(urls, sample_size, github_lang=None):
...
+    log_filename = os.path.join('output', 'generate.log')
```
inferno-chromium, line 99: we need to use args.output_dir instead of hardcoded output, it can be passed as arg to initialize_logging_handlers
also, generate.log can be renamed to output.log.

inferno-chromium (approved):
LGTM with minor changes needed.

criticality_score/generate.py
coni2k: Okay, it looks like this now:
https://github.com/ossf/criticality_score/pull/66/files

inferno-chromium (approved):

PR #67, opened 2021-01-09
Update format of generate script console output

coni2k:
When running generate script with large numbers, there's no way of knowing how many repos are processed and how many left to go. So, I wanted to display an index for each line. Since I started playing with it, I changed the output format a little.

**Current output**

```shell

name,url,language,created_since,updated_since,contributor_count,org_count,commit_frequency,recent_releases_count,updated_issues_count,closed_issues_count,comment_frequency,dependents_count,criticality_score
Hacky,https://github.com/eliaskg/Hacky,Objective-C,97,84,4,2,0.0,0,0,0,0,0,0.10185
flame.js,https://github.com/flamejs/flame.js,JavaScript,107,57,34,5,0.0,0,0,0,0,2,0.22108
```

**New output**: no header, each line only shows index, name, url, score

```shell
0 - iglo - https://github.com/subosito/iglo - 0.1077
1 - Hacky - https://github.com/eliaskg/Hacky - 0.10185
```

**Alternatively**, if we still want to see all the fields, I can keep the old format but only add the index like this:

```shell
name,url,language,created_since,updated_since,contributor_count,org_count,commit_frequency,recent_releases_count,updated_issues_count,closed_issues_count,comment_frequency,dependents_count,criticality_score
0 - Hacky,https://github.com/eliaskg/Hacky,Objective-C,97,84,4,2,0.0,0,0,0,0,0,0.10185
1 - flame.js,https://github.com/flamejs/flame.js,JavaScript,107,57,34,5,0.0,0,0,0,0,2,0.22108
```

Let me know what you think

### DIFF
--- a/criticality_score/generate.py
+++ b/criticality_score/generate.py
@@ -136,9 +136,8 @@ def main():
         logger.info(f'Finding repos (round {rnd}):')
         repo_urls.update(get_github_repo_urls(args.sample_size, args.language))
 
-    csv_writer = csv.writer(sys.stdout)
-    header = None
     stats = []
+    index = 1
     for repo_url in repo_urls:
         output = None
         for _ in range(3):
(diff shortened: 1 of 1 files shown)


inferno-chromium:
Sorry we cannot have this. csv format is needed to quickly import into google sheets, excel, etc and pretty standard way to do stuff. this new format does not help with that.

coni2k:
The output file itself still stays as csv. This change is only about the output to the console

inferno-chromium (approved):

inferno-chromium:
> The output file itself still stays as csv. This change is only about the output to the console

SG

PR #71, opened 2021-01-12
Refactor get_github_auth_token

coni2k:
I realized that it's possible to use multiple auth tokens for GitHub, which is a life-saver. Thanks for that ✌
Now generate script takes much less time to run.

While I was using multiple tokens, I noticed that `get_github_auth_token` function doesn't take the shortest wait time but just uses the wait time of the last token. Although not sure how much this would improve the process but I wanted to try refactoring it.

Here is the summary of this update:
- Introduced `_CACHED_GITHUB_TOKENS` global variable that stores `token`, `token_obj` and `reset` data as a list
- `get_github_auth_token` returns an object from this list
- While searching for the next active token, it ignores the already expired tokens
- If all tokens are expired, it finds the minimum wait time and goes to sleep
- That block is in a while loop and so it just goes back to start to find the next active token after the sleep

One more small update that I could do is to eliminate the use of `_CACHED_GITHUB_TOKEN` global variable, which is used in `get_first_commit_time` function. But that would mean one more "rate limit" check. If you think that's okay, I could update that part as well?

I keep sending these PRs to able to sync with my fork but since we don't have prior discussions about these updates, if it doesn't align with your work, feel free to ignore them as usual 🙏

### DIFF
--- a/criticality_score/run.py
+++ b/criticality_score/run.py
@@ -31,7 +31,7 @@ import requests
 from .constants import *  # pylint: disable=wildcard-import
 
 _CACHED_GITHUB_TOKEN = None
-_CACHED_GITHUB_TOKEN_OBJ = None
+_CACHED_GITHUB_TOKENS = []
 
 PARAMS = [
     'created_since', 'updated_since', 'contributor_count', 'org_count',
@@ -143,7 +143,8 @@ class GitHubRepository(Repository):
                     links[match.group(2)] = match.group(1)
             return links
 
-        headers = {'Authorization': f'token {_CACHED_GITHUB_TOKEN}'}
+        token = _CACHED_GITHUB_TOKEN['token']
+        headers = {'Authorization': f'token {token}'}
         for i in range(FAIL_RETRIES):
             result = requests.get(f'{self._repo.url}/commits', headers=headers)
             links = _parse_links(result)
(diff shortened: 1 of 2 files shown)


criticality_score/run.py
```diff
@@ -460,41 +461,54 @@ def _worker(repo, param, return_dict):
...
-          file=sys.stderr)
-    time.sleep(wait_time)
-    return token_obj
+
+    # Initialize _CACHED_GITHUB_TOKENS that stores token, token_obj and reset data as a list
+    if not _CACHED_GITHUB_TOKENS:
+        github_auth_token = os.getenv('GITHUB_AUTH_TOKEN')
+        assert github_auth_token, 'GITHUB_AUTH_TOKEN needs to be set.' # TODO Validate earlier?
+        for token in github_auth_token.split(','):
+            token_obj = github.Github(token)
```
inferno-chromium, line 480: We are recreating all token objects after every repo, this is not needed, also feels unnecessary to add reset value in there, since it needs to be rechecked. Also get_github_auth_token_info is weird now, it is just set the reset value, not getting any info.

If you want to just do the minimum wait time, you can do in the loop by capturing least value of wait_time while iterating in the loop. that is it, no other change needed.

    for token in tokens:	            
        token_obj = github.Github(token)
        near_expiry, wait_time = get_github_token_info(token_obj)
        if not near_expiry:	
            _CACHED_GITHUB_TOKEN = token	        
            _CACHED_GITHUB_TOKEN_OBJ = token_obj
            return token_obj

inferno-chromium (changes requested):

criticality_score/run.py
coni2k: Okay, only kept "min_wait_time" part and removed the other changes

inferno-chromium (approved):

PR #78, opened 2021-02-05
Fix regex to cover '1 commit result' case

Yikun:
Current regex doesn't cover '1 commit result' case, that means if we
get only 1 commit result, the dependent number will be set to 0.

This patch try to fix it by use '* commit result' to match 'n commit results'
and '1 commit result' case.

### DIFF
--- a/criticality_score/constants.py
+++ b/criticality_score/constants.py
@@ -45,4 +45,4 @@ RELEASE_LOOKBACK_DAYS = 365
 FAIL_RETRIES = 7
 
 # Regex to match dependents count.
-DEPENDENTS_REGEX = re.compile(b'.*[^0-9,]([0-9,]+).*commit results', re.DOTALL)
+DEPENDENTS_REGEX = re.compile(b'.*[^0-9,]([0-9,]+).*commit result', re.DOTALL)


Yikun:
Notes for reviewer:
```
regex_old=b'.*[^0-9,]([0-9,]+).*commit results'
regex_new=b'.*[^0-9,]([0-9,]+).*commit result'

# the old regex can't match 1 commit result
>>> re.compile(regex_old).match(b'<p>100 commit results</p>').group(1)
b'100'
>>> re.compile(regex_old).match(b'<p>1 commit result</p>').group(1)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'NoneType' object has no attribute 'group'

# the new regext can match 1 commit result and n commit results
>>> re.compile(regex_new).match(b'<p>100 commit results</p>').group(1)
b'100'
>>> re.compile(regex_new).match(b'<p>1 commit result</p>').group(1)
b'1'
```

[1] https://github.com/search?q=kunpengcompute%2Fkunpeng&type=commits

inferno-chromium (approved):

PR #79, opened 2021-02-06
Add page query retries when no match

Yikun:
When we do search on related commits of specific repo_name in github, we
perhaps get 200 status_code but no right page result, then get an
unexpected 0 result accidently.

This patch breaks retry loop only when result is 200 status with match
result rather than only 200 status.

### DIFF
--- a/criticality_score/run.py
+++ b/criticality_score/run.py
@@ -98,6 +98,7 @@ class Repository:
     def dependents_count(self):
         # TODO: Take package manager dependency trees into account. If we decide
         # to replace this, then find a solution for C/C++ as well.
+        match = None
         parsed_url = urllib.parse.urlparse(self.url)
         repo_name = parsed_url.path.strip('/')
         dependents_url = (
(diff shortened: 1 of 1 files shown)


criticality_score/run.py
```diff
@@ -109,7 +108,16 @@ def dependents_count(self):
                 content = result.content
                 break
             time.sleep(2**i)
-        match = DEPENDENTS_REGEX.match(content)
+        return DEPENDENTS_REGEX.match(content)
+
+    @property
+    def dependents_count(self):
+        # do query retries when no match to ensure accurate result
+        for i in range(FAIL_RETRIES):
```
inferno-chromium, line 116: We dont need two FAIL_RETRIES and a seperate function, can you modify the loop to not break on status 200. Break only when status is 200 and there is a match as well [and return directly from inside loop].

inferno-chromium (changes requested):

criticality_score/run.py
Yikun: Yep, done

inferno-chromium (approved):

PR #81, opened 2021-02-06
Fix pylint messages

coni2k:
Would you like to fix these four tiny pylint messages?


inferno-chromium (approved):
Thanks!
Probably should add a github actions for pylint

PR #84, opened 2021-02-10
Update Readme file to include "all.csv"

coni2k:


inferno-chromium (changes requested):
Since this does not use generation script, best to add it after line 150.

We have aggregated the results over 100K repositories in GitHub (language-independent) and are available for download [here](maybe web link or put gs://).

coni2k:
> Since this does not use generation script, best to add it after line 150.
> 
> We have aggregated the results over 100K repositories in GitHub (language-independent) and are available for download [here](maybe web link or put gs://).

Okay, updated with a link.

The link contains "generation" query parameter, which I assume that its sort of version of the file, so I kept it:
https://www.googleapis.com/download/storage/v1/b/ossf-criticality-score/o/all.csv?generation=1612987910088811&alt=media

inferno-chromium:
@coni2k - just fyi, i will be ooo tmrw, so can meet you in next-to-next wg meeting.

coni2k:
Sounds good, see you tomorrow then!

inferno-chromium:
> Sounds good, see you tomorrow then!

Sorry i meant i am out-of-office tmrw, so wont be attending tmrw meeting. But see you in the next one.

PR #86, opened 2021-02-21
Create Dependabot config file

naveensrinivasan:
Enable Dependabot for security notifications on dependencies https://dependabot.com/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+- package-ecosystem: pip
+  directory: "/"
+  schedule:
+    interval: daily
+  open-pull-requests-limit: 10


.github/dependabot.yml
```diff
@@ -0,0 +1,7 @@
+version: 2
+updates:
+- package-ecosystem: pip
```
inferno-chromium, line 3: Why pip ?

inferno-chromium (approved):

.github/dependabot.yml
naveensrinivasan: Sorry I use the dependabot to generate this. I don't know what package manager the repository uses. Is this wrong?

inferno-chromium: https://dependabot.com/docs/config-file/#package_manager-required
Should be these i think
go:modules
go:dep

naveensrinivasan: I don't see any package manager in this repository. Am I missing something?

inferno-chromium: dont know this. i only see docker container as package, dont see scorecard released as a go package (maybe we should). @dlorenc - can you help to review and see what is needed here.

inferno-chromium (changes requested):
Changes needed as per comment above.

.github/dependabot.yml
naveensrinivasan: Doesn’t this Python repo have dependency on pip packages?

inferno-chromium: I didn't see the repo, got confused with scorecards, yes criticality score is all pip.

inferno-chromium (approved):

PR #87, opened 2021-02-28
Handle get_tags exception

coni2k:
When we request `get_tags` for this repo, GitHub API throws an exception (interestingly, this was the only repo that was persistent in throwing an exception):
https://github.com/hortonworks/cloudbreak

```shell
python -u -m criticality_score.run --repo github.com/hortonworks/cloudbreak
```

It has 8,707 tags, so probably that's a large number for the API to process.

The only detail is that in catch block of "contributors", you return "10" as a default value. Should I set a similar default value for tags as well?

### DIFF
--- a/criticality_score/run.py
+++ b/criticality_score/run.py
@@ -259,7 +259,9 @@ class GitHubRepository(Repository):
             try:
                 total_tags = self._repo.get_tags().totalCount
             except Exception:
+                # Very large number of tags, i.e. 5000+. Cap at 26.
                 logger.error(f'get_tags is failed: {self._repo.url}')
+                return RECENT_RELEASES_THRESHOLD
             total = round(
                 (total_tags / days_since_creation) * RELEASE_LOOKBACK_DAYS)
         return total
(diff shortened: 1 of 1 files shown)


criticality_score/run.py
```diff
@@ -255,7 +255,11 @@ def recent_releases_count(self):
             days_since_creation = self.created_since * 30
             if not days_since_creation:
                 return 0
-            total_tags = self._repo.get_tags().totalCount
+            total_tags = 0
+            try:
+                total_tags = self._repo.get_tags().totalCount
+            except Exception:
+                logger.error(f'get_tags is failed: {self._repo.url}')
```
inferno-chromium, line 262: Add a comment that this failure is usually due to large number of tags.
Also, i think we should default to 26, or whatever that constant for max release count is (use constant from that file). this just means we default to max threshold which is equivalent to having any value larger than it (since it caps it).

inferno-chromium (changes requested):

criticality_score/run.py
coni2k: Okay, added the comment: `# Very large number of tags, i.e. 5000+. Cap at 26.`
And returning `RECENT_RELEASES_THRESHOLD` as the default value

inferno-chromium (approved):

PR #89, opened 2021-02-28
Update Readme

coni2k:
Update "all.csv" link in Readme


inferno-chromium (approved):
i  dont think you need that generation part, but fine either way.

PR #90, opened 2021-04-10
make github queries determinsitic

### LINKED ISSUE #33: Find workaround for github api limitation of 1000 repos in results, expand sample set beyond 1000 by stars before sorting by criticality score

tgamblin:
Fixes #33.

The current query method doesn't work well with GitHub's API because it makes queries that frequently return > 1,000 results.  For GitHub to be deterministic, you have to fetch < 1,000 results at a time.  This is not straightforward to do with stars, because stars obey a power law.  The lower the star minimum in a query, the more likely you are to get lots of results, because there are many repos with not lower numbers of stars.

This PR does things:
1. Try to keep the total number of API calls made to GitHub small.
2. Dynamically adjust the min and max stars we request via the GitHub API to contra the size of result sets.

To do (1), we start by requesting repos with > 65,536 stars, and initially we just halve that number each time through the fetch loop.  This rapidly ramps up the size of the query result sets.

To do (2), we also keep track of the size of result sets.  In particular, we look at how much bigger the last result set was than the one before, and we extrapolate how big the next one will be based on that.  It turns out that this is a good upper bound -- if you halve the minimum star count and the result set is k times larger than the one before, the following result set is typically < k times bigger than the most recent one.  If the expected number of results is too big (i.e. greater than a threshold) we reduce the divisor so that the next window is half as big as it would've been.

With these changes I can:
* Get repeatable results from GitHub
* Eliminate the 3 redundant rounds the generate code used to do
* Grab >5,000 results with ~14 API calls

I ran this a few times and got the same repositories in the result set each time.  I do see star counts change from query to query, so this may need to add a buffer (i.e. the star ranges in the queries should probably overlap by some fudge factor so we don't miss rapidly moving repos).  I didn't see any missed repos when I tried it, though.

Thanks to @mfridman and @coni2k for hints that led to this on #33.

@inferno-chromium what do you think?

### DIFF
--- a/criticality_score/generate.py
+++ b/criticality_score/generate.py
@@ -53,22 +53,69 @@ def get_github_repo_urls(sample_size, languages):
 def get_github_repo_urls_for_language(urls, sample_size, github_lang=None):
     """Return repository urls given a language list and sample size."""
     samples_processed = 1
-    last_stars_processed = None
-    while samples_processed <= sample_size:
 
+    star_min_start = 65536      # set high enough so that few repos have this many stars
+    star_min = star_max = None  # min and max stars in the window to fetch
+    divisor = 2                 # we start out halving the min for each successive query
+    threshold = 750             # set to be significantly lower than the limit of 1000
+
+    result_counts = []
+    while True:
         query = 'archived:false'
         if github_lang:
             query += f' language:{github_lang}'
 
-        if last_stars_processed:
-            # +100 to avoid any races with star updates.
-            query += f' stars:<{last_stars_processed+100}'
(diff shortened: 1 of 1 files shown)


inferno-chromium:
the reason for the repeated loops was also to cover race condition w.r.t star changes. while running the query, if stars change, you lose that repo altogether. for popular repo, it can be common, that is why i fixed a dozen where people found repos missing. i am fine with using more github api quota for more accurate results.

tgamblin:
@inferno-chromium: that can be fixed, as mentioned above, by making the queries overlap.  I’ll add that.

PR #95, opened 2021-08-04
Add Watchers/Description Metrics

dilanbhalla:
I wanted to submit a suggestion to include GitHub Watchers (to help assess popularity) and the GitHub Description (to clarify the project's overall goal). I am currently helping contribute to OSSF's Security Metrics project, in which we are retrieving several of the GitHub metrics covered in this project (but also need to analyze the two mentioned above to help with our overall security assessment). If these can be included via the pull request I have submitted that would be extremely helpful. Thank you!

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,17 +52,19 @@ $ criticality_score --repo github.com/kubernetes/kubernetes
 name: kubernetes
 url: https://github.com/kubernetes/kubernetes
 language: Go
-created_since: 79
+description: Production-Grade Container Scheduling and Management
+created_since: 87
 updated_since: 0
-contributor_count: 3664
+contributor_count: 3999
+watchers_count: 79583
 org_count: 5
-commit_frequency: 102.7
-recent_releases_count: 76
-closed_issues_count: 2906
-updated_issues_count: 5136
-comment_frequency: 5.7
-dependents_count: 407254
-criticality_score: 0.9862
+commit_frequency: 97.2
+recent_releases_count: 70
+updated_issues_count: 5395
+closed_issues_count: 3062
+comment_frequency: 5.5
+dependents_count: 454393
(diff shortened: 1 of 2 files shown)


inferno-chromium (approved):

dilanbhalla:
Thank you! Would it be possible to also refresh the Google Cloud Storage link to include this information @inferno-chromium?

dilanbhalla:
Hi @inferno-chromium, sure I can generate the files. A couple questions before doing so though.

An example of a command I have been running (as a small scale test) is `python -u -m criticality_score.generate --count 2 --sample-size 5 --output-dir output`, which seems to do the trick. When I want to generate the full output, can I simply run this command but set the count and sample size to 100,000?

Secondly, why does the link you posted on github also contain a "stars" and "license" column but the generator script from the command line does not? Is there a way to include this?

Thank you!

inferno-chromium:
> Hi @inferno-chromium, sure I can generate the files. A couple questions before doing so though.
> 
> An example of a command I have been running (as a small scale test) is `python -u -m criticality_score.generate --count 2 --sample-size 5 --output-dir output`, which seems to do the trick. When I want to generate the full output, can I simply run this command but set the count and sample size to 100,000?

try higher like 25K for both. i think across all languages might take 1-2 weeks. we want to expand scorecards list as well, so more repos will help here. 

> 
> Secondly, why does the link you posted on github also contain a "stars" and "license" column but the generator script from the command line does not? Is there a way to include this?

probably old code, those are not useful fields i think. can be excluded.

> 
> Thank you!

dilanbhalla:
Hi @inferno-chromium, I actually had a bunch of other work items pop up recently, but there is no rush for our team on this particular metric, so whenever the code is productionized and data is published, we will retrieve the information. Thanks!

PR #99, opened 2021-10-22
Normalize scores to handle negative weight

Jehops:
Because of the negative weight used for the 'updated_since' signal,
calculated criticality scores can take on values outside of the desired
range of [0,1].  This was recognized and a workaround was committed in
cae562b0e0dd0bf9a043026a77b11f6fd1e1138a.

criticality_score = max(min(criticality_score, 1), 0)

In words, set the score to 1 if the calculated score is greater than 1
and set the score to 0 if the calculated score is less than 0.  That
workaround removes the convenient interpretation of a score as the
proportion of the maximum score.  It also removes some information.  For
example, we miss that a score of 1.03 before the workaround is applied
is better than a score of 1.00.

The proposed solution here restores the desired range, preserves the
interpretation of a score as the proportion of the maximum score, and
does not throw out any information.

### DIFF
--- a/criticality_score/run.py
+++ b/criticality_score/run.py
@@ -439,7 +439,8 @@ def get_repository_stats(repo, additional_params=None):
         return None
     if additional_params is None:
         additional_params = []
-    additional_params_total_weight = 0
+    additional_params_max_score = 0
+    additional_params_min_score = 0
     additional_params_score = 0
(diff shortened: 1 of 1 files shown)


calebbrown:
Hi, the Python code has been deprecated and support for handling negative weights has been incorporated into the new Go implementation.

Issue #104, opened 2022-03-31
Overriding the default parameters


lehors:
The weight and threshold associated with the various parameters involved in the criticality score are to some degree arbitrary and an organization may want to change those to better fit its own needs.
There should therefore be a way to override those values at runtime.

And I have a PR to do just that. :-)

PR #105, opened 2022-03-31
Allow overriding default parameters

lehors:
Adds a command line parameter to override the weight and threshold
of the build-in parameters.

This addresses Issue #104 

Signed-off-by: Arnaud J Le Hors <lehors@us.ibm.com>

### DIFF
--- a/criticality_score/run.py
+++ b/criticality_score/run.py
@@ -29,7 +29,7 @@ import github
 import gitlab
 import requests
 
-from .constants import *  # pylint: disable=wildcard-import
+from .defaults import *  # pylint: disable=wildcard-import
 
 logger = logging.getLogger()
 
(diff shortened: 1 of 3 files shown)


criticality_score/run.py
```diff
@@ -608,6 +608,62 @@ def initialize_logging_handlers():
     logging.getLogger('').addHandler(console)
 
 
+def override_params(override_params):
```
calebbrown, line 611: Would it make more sense to restructure `constants`/`defaults` to make it easier to replace the default values, rather than mutating the initial values?

e.g. turn the constants into a dict, where each value is a namedtuple or dataclass and pass the dict (with any overrides) as an argument into the scoring method.

criticality_score/run.py
```diff
@@ -608,6 +608,62 @@ def initialize_logging_handlers():
...
+    for override_param in override_params:
+        temp = override_param.split(':',1)
+        param_name = temp[0]
+        try:
+            weight, threshold = [
+                int(i) for i in temp[1].split(':')
```
calebbrown, line 617: Is it worth supporting `float` here? Particularly for `weight`.

calebbrown (commented):
Thanks for this change - changing weights and thresholds is certainly useful for experimenting with the score.


FYI: See #108 on the future direction I have in mind for this project.

criticality_score/run.py
```diff
@@ -627,10 +683,18 @@ def main():
         default=[],
         help='Additional parameters in form <value>:<weight>:<max_threshold>',
         required=False)
+    parser.add_argument(
+        '--overrides',
```
calebbrown, line 687: Is there any sense in combining `--overrides` with `--params`?

criticality_score/run.py
lehors: There always are many ways one can implement things and, especially when I'm new to a project, I try to keep my PRs to a minimum to make them easier to be accepted. I thought it'd be argued that rather than renaming constants.py defaults.py I should split the file in two and only put in defaults.py things that can indeed be modified. :-)

criticality_score/run.py
lehors: That's a good question! We've just started playing with this and all the defaults were integers so that's what I went with but it might be useful to make them floats indeed. We don't have enough experience to say whether we'd make use of that at this point but if you think that's worth doing I'm happy to change it to float.

criticality_score/run.py
lehors: I've actually thought that --params should support specifying a name and the results should display those along with the built-in parameters (they currently aren't), if we did so we could combine the two with a syntax such as:
`--params <name>:<weight>:<max_threshold>[:<value>]` specifying that if the name matches a built-in param it is an override.
On the other hand this would make the command usage more complicated to understand. The separate option has the advantage of being more explicit.

calebbrown: Agreed. Happy to leave it separate

criticality_score/run.py
calebbrown: Some of the weights are `floats` already so I think it makes sense to do it for the weights at least

criticality_score/run.py
calebbrown: In this case I don't mind the extra effort to improve the overall implementation. But I'm happy to approve this change as is as it doesn't change the things in a way that can't be changed to a `dict` based approach in a future PR.

criticality_score/run.py
lehors: You're right, I changed it to float.

calebbrown (approved):

PR #111, opened 2022-04-12
Add a command for enumerating GitHub repos (written in Go)

calebbrown:
Takes the algorithms suggested in #33 and combines them to allow approx
repos with at least 10 stars and higher to be enumerated.

This approach *only* collects project URLs.

There are other sources of data that can be used such as GH Archive,
GHTorrent, etc to compliment this list.

All the project URLs found can eventually be de-duped, checked for
existance (e.g. 404s), canonicalized, and then used for signal collection.


cmd/enumerate_github/main.go
```diff
@@ -0,0 +1,238 @@
...
+func searchWorker(s *githubsearch.Searcher, wg *sync.WaitGroup, logger logr.Logger, queries, results chan string) {
+	defer wg.Done()
+	for q := range queries {
+		total := 0
+		err := s.ReposByStars(q, *minStarsFlag, *starOverlapFlag, func(repo string) {
+			results <- repo
+			total++
+		})
+		if err != nil {
+			// TODO: this error handling is not at all graceful, and hard to recover from.
```
azeemshaikh38, line 109: Could consider having an `err` field in the `result` struct. In case of failure, this field will get populated and the error handling can happen after the go routines finish.

azeemshaikh38 (approved):

oliverchang (approved):
nice!

cmd/enumerate_github/main.go
```diff
@@ -0,0 +1,238 @@
...
+func main() {
+	flag.Parse()
+
+	// roundtripper requires us to use the scorecard logger.
+	// TODO: try and replace this with logrus directly.
+	logger := log.NewLogger(logFlag.Level())
+
+	// Ensure the -start date is not before the epoch.
+	if startDateFlag.Time().Before(epochDate) {
+		fmt.Fprintf(os.Stderr, "-start date must be no earlier than %s\n", epochDate.Format(githubDateFormat))
```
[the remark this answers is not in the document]
calebbrown: Fixed! Thanks

cmd/enumerate_github/main.go
calebbrown: Thanks for the suggestion. I haven't yet figured out exactly what I want to do here.

I plan to address this is a future PR.

PR #120, opened 2022-05-20
Initial commit of the collect_signals command.

calebbrown:
- defines key APIs:
  - projectrepo.Repo and projectrepo.Factory
  - signal.Set and signal.Signal
  - collector.Collector and collector.Registry
- implements the complete current set of signals for GitHub repos.
- supports CSV output.
- refactors GitHub API related code into an internal package for reuse.
- requires Go 1.18 for fancy features like generics and `any`
- clarified the field types in the milestone 1 doc, particularly around nullness and date/time formats.

Not yet there:
- documentation needs extending
- cmd/collect_signals/github/queries.go is a bit of a mess
- more unit tests


calebbrown:
Sorry for the *huge* commit.

internal/githubapi/client.go
```diff
@@ -0,0 +1,30 @@
+package githubapi
+
+import (
+	"net/http"
+
+	"github.com/google/go-github/v44/github"
+	"github.com/shurcooL/githubv4"
+)
+
+type Client struct {
```
naveensrinivasan, line 10: The consumers are directly using the GitHub clients. This is going to make it harder to unit test. The suggestion would be to use an interface similar to this https://github.com/ossf/scorecard/blob/7ac81a334f7ef5281d3238ffd68499a6a48106a8/clients/repo_client.go#L29 https://github.com/ossf/scorecard/blob/7ac81a334f7ef5281d3238ffd68499a6a48106a8/clients/repo.go#L18

naveensrinivasan (commented):
Thank you!

cmd/collect_signals/collector/collector.go
```diff
@@ -0,0 +1,51 @@
...
+	Collect(context.Context, projectrepo.Repo) (signal.Set, error)
+}
+
+// A LimitedCollector is a Collector that will only gather a Set of signals
+// for a supported repository.
+type LimitedCollector interface {
+	Collector
+
+	// IsSupported returns true if the Collector supports the supplied Repo.
+	IsSupported(projectrepo.Repo) bool
```
oliverchang, line 30: Question: Why do we need the distinction between `Collector` and `LimitedCollector`? 

Why not just require every Collector to implement `IsSupported` ?

cmd/collect_signals/collector/registry.go
```diff
@@ -0,0 +1,128 @@
...
+		panic(err)
+	}
+	r.cs = append(r.cs, &collectorWrapper{Collector: c})
+}
+
+func (r *Registry) collectorsForRepository(repo projectrepo.Repo) []Collector {
+	// Check for duplicates using a map to preserve the insertion order
+	// of the collectors.
+	exists := make(map[signal.Namespace]struct{})
+	res := make([]Collector, 0)
```
oliverchang, line 52: nit: per https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices I think the preferred way to do this is `var res []Collector`?

cmd/collect_signals/collector/registry.go
```diff
@@ -0,0 +1,128 @@
...
+		}
+		ss = append(ss, c.EmptySet())
+	}
+	return ss
+}
+
+// Collect will collect all the signals for the given repo.
+func (r *Registry) Collect(ctx context.Context, repo projectrepo.Repo) ([]signal.Set, error) {
+	cs := r.collectorsForRepository(repo)
+	ss := []signal.Set{}
```
oliverchang, line 88: same here: https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

cmd/collect_signals/collector/registry.go
```diff
@@ -0,0 +1,128 @@
...
+	return res
+}
+
+// EmptySets returns all the empty signal Sets for all the registered
+// Collectors.
+//
+// This result can be used to determine all the signals that are defined.
+func (r *Registry) EmptySets() []signal.Set {
+	exists := make(map[signal.Namespace]struct{})
+	ss := make([]signal.Set, 0)
```
oliverchang, line 74: same here: https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

cmd/collect_signals/github/queries.go
```diff
@@ -0,0 +1,75 @@
...
+	} `graphql:"refs(refPrefix:\"refs/tags/\")"`
+}
+
+func queryBasicRepoData(ctx context.Context, client *githubv4.Client, u *url.URL) (*basicRepoData, error) {
+	// Search based on owner and repo name becaues the `repository` query
+	// better handles changes in ownership and repository name than the
+	// `resource` query.
+	parts := strings.Split(strings.Trim(u.Path, "/"), "/")
+	owner := parts[0]
+	name := parts[1]
```
oliverchang, line 61: Do we need to handle ".git" suffixes here?

cmd/collect_signals/main.go
```diff
@@ -0,0 +1,191 @@
...
+			logger.WithFields(log.Fields{
+				"error":    err,
+				"filename": inFilename,
+			}).Error("Failed to open an input file")
+			os.Exit(2)
+		}
+		defer f.Close()
+		readers = append(readers, f)
+	}
+	r := io.MultiReader(readers...)
```
oliverchang, line 87: nit: this function is very long. Does it make senes to break up this reader aggregation part into its own helper function and call it here instead?

Other parts of this function can probably be broken up too.

cmd/collect_signals/result/csv.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func marshalValue(value any) (string, error) {
+	switch v := value.(type) {
+	case bool, int, int16, int32, int64, uint, uint16, uint32, uint64, byte, float32, float64, string:
+		return fmt.Sprintf("%v", value), nil
+	case time.Time:
+		return v.Format(time.RFC3339), nil
+	default:
+		return "", nil
```
oliverchang, line 94: should this be an error instead?

cmd/collect_signals/main.go
```diff
@@ -0,0 +1,191 @@
...
+	rt := roundtripper.NewTransport(ctx, scLogger)
+	httpClient := &http.Client{
+		Transport: rt,
+	}
+	ghClient := githubapi.NewClient(httpClient)
+
+	ghRepoFactory := github.NewRepoFactory(ghClient, logger)
+
+	// Register all the collectors that are supported.
+	collector.Register(&github.RepoCollector{})
```
oliverchang, line 113: Would it make sense to do these register calls in the init() of the relevant packages instead?

cmd/collect_signals/github/legacy/contributors.go
```diff
@@ -0,0 +1,108 @@
...
+	if len(userQueries) == 0 {
+		// We didn't add any users.
+		return 0, err
+	}
+	r, err := githubapi.BatchQuery[struct{ Company string }](ctx, c, userQueries, map[string]any{})
+	if err != nil {
+		return 0, err
+	}
+	// Extract the Company from each returned field and add it to the org set.
+	orgSet := make(map[string]struct{})
```
oliverchang, line 85: nit: define an `empty` type for readability?

oliverchang (commented):
some initial comments! still trying to digest the other bits (especially the generics/signal.Set parts).

cmd/collect_signals/github/queries.go
```diff
@@ -0,0 +1,75 @@
...
+	s := &struct {
+		Repository basicRepoData `graphql:"repository(owner: $repositoryOwner, name: $repositoryName)"`
+	}{}
+	now := time.Now().UTC()
+	vars := map[string]interface{}{
```
oliverchang, line 66: nit: use `any` instead of `interface{}` ? there's a number of other places in this PR.

cmd/collect_signals/signal/signal.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+// valuer is provides access to the field's value without needing to use
+// the concrete generic. It is also a marker interface used to identify the
+// Signal field in a struct.
+type valuer interface {
+	Value() any
+}
+
+type Field[T SupportedType] struct {
```
oliverchang, line 63: More general question as someone very unfamiliar with Go generics: there's a lot of reflection used here to make this package work. 

What value does using generics give us here over just purely using reflection? Is it purely to make this easier to implement? 

The signal sets current require sets to be defined as:

```
type IssuesSet struct {
	UpdatedCount     Field[int]     `signal:"updated_issues_count,legacy"`
	ClosedCount      Field[int]     `signal:"closed_issues_count,legacy"`
	CommentFrequency Field[float64] `signal:"issue_comment_frequency,legacy"`
}
```

instead of just:

```
type IssuesSet struct {
	UpdatedCount    int     `signal:"updated_issues_count,legacy"`
	ClosedCount      int     `signal:"closed_issues_count,legacy"`
	CommentFrequency float64 `signal:"issue_comment_frequency,legacy"`
}
```

Which adds a layer of indirection -- what does this indirection buy us? If the goal is to just be able to extract key/value pairs from this, we could just do something similar to what encoding/json is doing right?

calebbrown: The problem is that I want to support the idea of a field being "unset".

Using normal scalars like `int` is impossible for this. I could then use `*int`, and define my own `FieldInt` type, but then I might as well start using generics, e.g: `Field[T]` as I have.

The alternative is to allow `*int` and `int`, etc, and make the code handle it appropriately, but I suspect it is then easier to accidentally use an `int` and forget about the "unset" scenario. Trying to enforce pointer usage could only be done at runtime using reflection.

I feel that by using generics I can limit the use of reflection, and lean on the compiler to do a lot of the heavy lifting w.r.t type checking. I can also add functions to the `Field[T]` type as well that help with processing the data.

oliverchang: Thanks, that makes sense. I didn't think about this supporting the "unset" case.

cmd/collect_signals/main.go
calebbrown: I considered that.

In some cases a collector will need parameters passed in - and I think it will be easier to see a full set of collectors. Additionally the order of calls to `Register()` is preserved in the output, which can't be guaranteed with `init()` so much.

cmd/collect_signals/github/queries.go
calebbrown: Done.

cmd/collect_signals/collector/collector.go
calebbrown: Good idea. Done

cmd/collect_signals/collector/registry.go
calebbrown: Done.

cmd/collect_signals/collector/registry.go
calebbrown: Done.

cmd/collect_signals/main.go
oliverchang: ack.

calebbrown: FWIW I've added some more docs around the ordering.

`Collect()` will get some too soon, after I've re-implemented it to use goroutines.

cmd/collect_signals/result/csv.go
calebbrown: Probably a good idea. Done.

cmd/collect_signals/github/collector.go
```diff
@@ -0,0 +1,120 @@
...
+	if up, err := legacy.FetchIssueCount(ctx, ghr.client, ghr.owner(), ghr.name(), legacy.IssueStateAll, legacy.IssueLookback); err != nil {
+		return nil, err
+	} else {
+		s.UpdatedCount.Set(up)
+		if up != 0 {
+			ghr.logger.Debug("Fetching comment frequency")
+			if comments, err := legacy.FetchIssueCommentCount(ctx, ghr.client, ghr.owner(), ghr.name(), legacy.IssueLookback); err != nil {
+				if errors.Is(err, legacy.TooManyResultsError) {
+					ghr.logger.Debug("Comment count failed with too many result")
+					s.CommentFrequency.Set(2.0)
```
oliverchang, line 96: nit: put 2.0 in a constant to make it clear this is the default? 

Also, why 2.0 as the default?

cmd/collect_signals/github/legacy/issues.go
```diff
@@ -0,0 +1,62 @@
...
+func FetchIssueCommentCount(ctx context.Context, c *githubapi.Client, owner, name string, lookback time.Duration) (int, error) {
+	since := time.Now().UTC().Add(-lookback)
+	opts := &github.IssueListCommentsOptions{
+		Since:       &since,
+		ListOptions: github.ListOptions{PerPage: 1}, // 1 result per page means LastPage is total number of records.
+	}
+	cs, resp, err := c.Rest().Issues.ListComments(ctx, owner, name, 0, opts)
+	// The API returns 5xx responses if there are too many comments.
+	if c := githubapi.ErrorResponseStatusCode(err); 500 <= c && c < 600 {
+		return 0, TooManyResultsError
```
oliverchang, line 53: This seems inconsistent with the other implementations' behaviour (e.g. contributor count), which seems to hide this detail from the caller and just return the maximum number. 

i.e. 

```go
	cs, resp, err := c.Rest().Repositories.ListContributors(ctx, owner, name, opts)
	if errorTooManyContributors(err) {
		return MaxContributorLimit, nil
	}
```

Would that make more sense here too?

oliverchang (approved):
LGTM with remaining comments addressed. Thanks!

cmd/collect_signals/main.go
calebbrown: Agreed.

For now though I'd prefer to leave it as-is. Moving the reader logic to a separate function complicates the file closing behaviour.

cmd/collect_signals/github/collector.go
calebbrown: 2.0 was roughly what I was seeing on repos (usually between 1.5 and 2.5) for comment frequency.

It is a bit of a magic value, and leaving it unset may be more accurate.

cmd/collect_signals/github/legacy/issues.go
calebbrown: I'm not a huge fan of this difference, but it currently exists this way due to the usage of the results from `FetchIssueCount()` to generate the frequency (i.e. average comments per issue).

I have an refactoring in mind to improve the API around a struct{} that could also be used to reduce the number of args to the methods and provide some caching - making it possible to move this into a more logical location.

For the time being I'd be happy to get this merged and do some future refactoring soon (I've added a TODO).

oliverchang: ack.

cmd/collect_signals/github/collector.go
oliverchang: ack.

oliverchang (approved):

internal/githubapi/client.go
calebbrown: Thanks for the feedback, interfaces certainly help with testability and I do need to do some work in this area.

At the moment I can mock the HTTP libraries, but that may not be the most practical approach.

I expect this area to evolve significantly over the next few months.

PR #125, opened 2022-06-01
Add support for multiple workers to the signal collection

calebbrown:


cmd/collect_signals/main.go
```diff
@@ -44,6 +50,71 @@ func init() {
...
+			logger.WithFields(log.Fields{
+				"status":      arle.Response.Status,
+				"headers":     arle.Response.Header,
+				"message":     arle.Message,
+				"retry_after": arle.RetryAfter,
+			}).Error("Abuse Error error...")
+		}
+		var er *gh.ErrorResponse
+		if errors.As(err, &er) {
+			f, err2 := os.Create("fail.html")
```
oliverchang, line 82: nit: do we actually need `err2`? We don't need the original `err` in this scope right?

cmd/collect_signals/main.go
```diff
@@ -134,58 +218,20 @@ func main() {
...
+		repos <- u
 	}
 	if err := scanner.Err(); err != nil {
 		logger.WithFields(log.Fields{
 			"error": err,
 		}).Error("Failed while reading input")
 		os.Exit(2)
 	}
+	// Close the repos channel to indicate that there is no more input.
+	close(repos)
```
oliverchang, line 231: Should this be in a defer instead?

oliverchang (approved):
nice!

cmd/collect_signals/main.go
calebbrown: No. It needs to happen before `wait()` defer would force it to happen afterwards.

At some point I will need to gracefully terminate other workers on a SIGTERM or a dead worker (unrecoverable error), but that will need more improvements.

cmd/collect_signals/main.go
calebbrown: This was debugging code. I've removed it.

PR #126, opened 2022-06-03
Add a library to handle the retry logic.

calebbrown:
A following PR will use this library to handle retries for the GitHub API.

This library is written this way to facilitate testing, which I thought
was particularly important for retry related logic as it is hard to
cover all the various cases during manual testing.

### DIFF
--- a/internal/retry/request.go
+++ b/internal/retry/request.go
@@ -146,8 +146,25 @@ func NewRequest(r *http.Request, client func(*http.Request) (*http.Response, err
 	}
 }
 
+// Done indicates whether or not the request can be attempted any more times.
+//
+// Initially Done will be false. It will move to true when the following
+// occurs:
+//
+//  1. Do returns a 2xx or 3xx response
+//  2. Do returns an error
+//  3. The number of attempts exceeds MaxRetries
+//  4. No RetryStrategy was returned or only NoRetry, and RetryAfter() had no
+//  delay.
+//
+// If Done returns true, Do must never be called again, otherwise Do will
+// return ErrorNoMoreAttempts.
+//
+// If Done returns false, Do needs to be called.
+//
+// If Do has never been called, this method will always return false.
 func (r *Request) Done() bool {
-	return r.done || r.attempts >= r.o.maxRetries
+	return r.done || r.attempts > r.o.maxRetries
 }
 
 func (r *Request) onError(err error) (*http.Response, error) {
--- a/internal/retry/request.go
+++ b/internal/retry/request.go
@@ -1,0 +1,248 @@
+package retry
+
+import (
+	"errors"
+	"net/http"
+	"time"
+)
+
+const (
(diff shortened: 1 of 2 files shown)


internal/retry/request.go
```diff
@@ -0,0 +1,248 @@
...
+type roundTripper struct {
+	inner http.RoundTripper
+	o     *Options
+}
+
+func (rt *roundTripper) RoundTrip(r *http.Request) (*http.Response, error) {
+	var resp *http.Response
+	var err error
+	rr := NewRequest(r, rt.inner.RoundTrip, rt.o)
+	for !rr.Done() {
```
oliverchang, line 244: in what case would this be done already?

internal/retry/request.go
```diff
@@ -0,0 +1,248 @@
...
+func NewRequest(r *http.Request, client func(*http.Request) (*http.Response, error), o *Options) *Request {
+	return &Request{
+		client: client,
+		r:      r,
+		o:      o,
+	}
+}
+
+func (r *Request) Done() bool {
+	return r.done || r.attempts >= r.o.maxRetries
```
oliverchang, line 150: does `attempts` include the first attempt? If so, semantically this should be `attempts > maxRetries` right?

internal/retry/request.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+
+	// Immediate success if the request is 2xx or 3xx
+	if http.StatusOK <= resp.StatusCode && resp.StatusCode < http.StatusBadRequest {
+		return r.onDone(resp, err)
+	}
+
+	if r.o.retryAfter != nil {
+		// Check if the Retry-After header is set
+		d := r.o.retryAfter(resp)
```
oliverchang, line 195: just curious: Why do we get the user to provide this function?

oliverchang (approved):

internal/retry/request.go
calebbrown: Never, immediately after instantiation.

The forloop semantics are convenient for being able to make the retry library testable.

Technically you could implement this as:

```go
for {
  resp, err := rr.Do()
  if rr.Done() {
    break
  }
}
```

I'll add a comment making this clearer.

internal/retry/request.go
calebbrown: Semantically yes. And I will change this to match.

This means `Options.maxRetries == 0` always results in a single attempt, but no retry.

internal/retry/request.go
calebbrown: Mostly for logging reasons at the moment (the current location has logging support). I'll add a TODO there to move it here later when I'm more confident in its correctness.

internal/retry/request.go
oliverchang: ack. I misread the code here actually. thanks!

PR #127, opened 2022-06-03
Add retry logic for github errors to improve reliability.

calebbrown:
1. Adds a RetryAfter function for processing responses with `Retry-After` header (will move to the retry library in the future)
2. Adds a SecondaryRateLimit function for detecting rate limiting from GitHub and pausing immediately
3. Adds a ServerError400 function for detecting odd server errors from GitHub
4. Adds a ServerError function for detecting any 5xx server error and retrying (except for `issues/comments` which knowingly fails on lots for large numbers of comments).

All of these are well tested to ensure the retry behavior works correctly.

### DIFF
--- a/internal/githubapi/roundtripper.go
+++ b/internal/githubapi/roundtripper.go
@@ -100,11 +100,11 @@ func (s *strategies) SecondaryRateLimit(r *http.Response) (retry.RetryStrategy, 
 			}).Warn("ReadAll failed.")
 		return retry.NoRetry, err
 	}
-	// Don't error check the unmarshall - if it is an error and the parsing has
-	// failed then will return with no retry. A parsing error here would mean
-	// the server did something wrong. Not attempting a retry will cause the
-	// response to be processed again by go-github with an error being
-	// generated there.
+	// Don't error check the unmarshall - if there is an error and the parsing
+	// has failed then this function will return with no retry. A parsing error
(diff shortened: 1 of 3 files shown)


calebbrown:
For reference: I have a job still running at nearly 72 hours non-stop on 1 token with this error handling, so it appears to be sufficient.

internal/githubapi/roundtripper.go
```diff
@@ -0,0 +1,129 @@
...
+	r.Body = ioutil.NopCloser(bytes.NewBuffer(data))
+	if err != nil || data == nil {
+		s.logger.WithFields(
+			log.Fields{
+				"error":    err,
+				"data_nil": (data == nil),
+			}).Warn("ReadAll failed.")
+		return retry.NoRetry, err
+	}
+	json.Unmarshal(data, errorResponse)
```
oliverchang, line 103: Should we be checking for error here?

oliverchang (approved):

internal/githubapi/roundtripper.go
calebbrown: This code is copied from the go-github repo. If it's not valid JS then this response should fall through to the other error handling.

calebbrown: Added a comment to make this clearer.

oliverchang: ack, thanks!

PR #131, opened 2022-06-14
Initial import of a scorer command for scoring a csv file of signals.

calebbrown:
- loads a YAML config file that defines the fields, weights, bounds, etc.
- scores each row of signals as generated by `collect_signals`, adding the result to a new column
- sorts the records and outputs it as a CSV

The config attempts to balance the flexibility of adding/removing fields and adjusting weights, with the complexity of the config format.

TODO: some unit testing and documentation


cmd/scorer/algorithm/input.go
```diff
@@ -0,0 +1,51 @@
...
+		v = b.Lower
+	} else if v > b.Upper {
+		v = b.Upper
+	}
+	// Afterwards we move L to 0, by calculating v = v - L
+	v = v - b.Lower
+	if b.SmallerIsBetter {
+		v = b.Threshold() - v
+	}
+	// If "SmallerIsBetter" is true, then we move the value
```
oliverchang, line 23: should this comment be moved up a bit?

cmd/scorer/algorithm/registry.go
```diff
@@ -0,0 +1,53 @@
...
+// If another Factory has been registered with the same Key it will be
+// replaced.
+func (r *Registry) Register(name string, f Factory) {
+	r.as[name] = f
+}
+
+// NewAlgorithm generates a new instance of Algorithm for the supplied name and
+// fields.
+//
+// If the registery does not have a Factory for the supplied name an error will
```
oliverchang, line 32: nit: spelling of registry.

config/scorer/pike_depsdev.yml
```diff
@@ -0,0 +1,84 @@
...
+    distribution: zapfian
+
+  # If deps.dev dependenct count doesn't exist we use this configuration
+  # for the GitHub search mention count.
+  - field: legacy.github_mention_count
+    weight: 2
+    bounds:
+      upper: 500000
+    condition:
+      not: yes
```
oliverchang, line 66: nit: perhaps this would be clearer as 

```yaml
condition:
  not:
    field_exists: "depsdev.dependent_cuont"
```

`not: yes` is a bit confusing at first glance.

oliverchang (approved):
nice!! just some minor comments.

cmd/scorer/main.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+	// Get the name of the config file used, without the path
+	f := path.Base(*configFlag)
+	ext := path.Ext(f)
+	// Strip the extension and convert to lowercase
+	f = strings.ToLower(strings.TrimSuffix(f, ext))
+	// Change any non-alphanumeric character into an underscore
+	f = regexp.MustCompile("[^a-z0-9_]").ReplaceAllString(f, "_")
+	// Append "_score" to the end
+	return fmt.Sprintf("%s_score", f)
```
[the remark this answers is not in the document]
calebbrown: Done

cmd/scorer/algorithm/input.go
calebbrown: Done

cmd/scorer/algorithm/registry.go
calebbrown: Fixed

cmd/scorer/main.go
```diff
@@ -0,0 +1,231 @@
...
+		os.Exit(2)
+	}
+	if err := w.Write(outHeader); err != nil {
+		logger.WithFields(log.Fields{
+			"error": err,
+		}).Error("Failed to write CSV header row")
+		os.Exit(2)
+	}
+
+	var pq PriorityQueue
```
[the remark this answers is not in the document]
calebbrown: I agree in principal that this would be better outside of this tool, but the ergonomics of `sort` isn't that great.

`sort` doesn't handle the header row nicely, and the column is hard to specify.

I'm going to leave this in for now. I have in mind preparing a collection of tools for analysing the csv data (e.g. stats like CDF, median, min, max, correlation, and the differences in ranking between two sets of scores) and this could exist alongside it.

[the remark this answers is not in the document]
oliverchang: ack!

config/scorer/pike_depsdev.yml
calebbrown: True. I've implemented this. In the process I've built something that could easily be extended to support "and:" and "or:" and other arbitrary boolean conditional logic (e.g. field_a > field_b).

oliverchang: Nice! Let's try to avoid other complicated logical operators if we can though :)

oliverchang (approved):

PR #134, opened 2022-06-17
Fix a bug where the params where set in the query.

calebbrown:


oliverchang (approved):

calebbrown:
Without this the BQ crashes because the "projectname" and "projecttype" parameters are not set.

PR #172, opened 2022-08-25
Add copyright notice to all files

calebbrown:
All files need to have a copyright notice.


oliverchang (commented):
Not sure this is right. Look at e.g. https://github.com/ossf/scorecard/blob/main/pkg/common.go where we say "The Security Scorecard Authors"

oliverchang (approved):

PR #179, opened 2022-08-31
Add an initial k8s config for running the enumerate_github tool.

calebbrown:
Currently, only runs for repos >= 100 stars, for 2021.

It is triggered each day at 23:00 UTC or 9am AEST.

Prior to running it needs the bucket to be created with access for the correct service account.


oliverchang (approved):
Awesome!

PR #228, opened 2022-11-03
Make the output file an explicit flag, rather than a positional arg

calebbrown:
Use -out=file to specify the output file, rather than as a positional arg.

This change is important for simplifying the CLI for `criticality_score`, so it was applied across all the commands.

The following changes were made to support this:
- A flag was added to outfile for the filename
- The capability to transform filenames was added to outfile (to support run-ids in enumeration)
- All the READMEs were updated.

This changes moves us closer to deprecating the Python version

### DIFF
--- a/cmd/criticality_score/README.md
+++ b/cmd/criticality_score/README.md
@@ -12,8 +12,8 @@ $ export GITHUB_TOKEN=ghp_x  # Personal Access Token Goes Here
 $ gcloud login --update-adc  # Sign-in to GCP
 $ criticality_score \
     -workers=1 \
-    github_projects.txt \
-    signals.csv
+    -out=signals.csv \
+    github_projects.txt
 ```
 
 ## Install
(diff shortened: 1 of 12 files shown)


cmd/criticality_score/README.md
```diff
@@ -25,14 +25,14 @@ $ go install github.com/ossf/criticality_score/cmd/criticality_score
...
-$ criticality_score [FLAGS]... IN_FILE OUT_FILE
+$ criticality_score [FLAGS]... IN_FILE
 ```
 
 Project repository URLs are read from the specified `IN_FILE`. If `-` is passed
 in as an `IN_FILE` URLs will read from STDIN.
 
-Results are written in CSV format to `OUT_FILE`. If `OUT_FILE` is `-` the
-results will be written to STDOUT.
+Results are written in CSV format the output. By default `stdout` is used for
```
oliverchang, line 34: Fix grammar of the first sentence?

oliverchang (approved):

oliverchang (approved):

PR #229, opened 2022-11-03
Change how repo URL are input into criticality_score + and create a docker image

calebbrown:
This PR changes the API to criticality_score to make it easy for use from the command line.

Repos can now be specified from the command line, rather than depending on piping input from echo. This makes it much easier for someone to use the tool without having to install it.

This change also introduces a Docker image for the criticality_score. By publishing a docker image regularly people will be able to use the project just by using the docker image (including alternative scoring configurations).

After this change is merged Python deprecation is now straight-forward.

### DIFF
--- a/cmd/criticality_score/input.go
+++ b/cmd/criticality_score/input.go
@@ -1,0 +1,139 @@
+// Copyright 2022 Criticality Score Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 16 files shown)


cmd/criticality_score/input.go
```diff
@@ -0,0 +1,139 @@
...
+// If more than one arg is specified they are all considered to be repos.
+//
+// TODO: support the ability to force args to be interpreted as either a file,
+// or a list of repos.
+func initInput(args []string) (iterCloser[string], error) {
+	if len(args) == 1 {
+		// If there is 1 arg, attempt to open it as a file.
+		fileOrRepo := args[0]
+		_, err := url.Parse(fileOrRepo)
+		notAUrl := err != nil
```
oliverchang, line 118: nit: this should be `notAURL` per https://github.com/golang/go/wiki/CodeReviewComments#initialisms. Or if that looks ugly maybe we can just invert and call this `isFilePath` or something.

oliverchang (approved):
nice!

cmd/criticality_score/input.go
calebbrown: lol. Style rule gets me every time.

oliverchang (approved):

PR #230, opened 2022-11-07
Implement the collect_signals worker for productionization.

calebbrown:
The `collect_signals` command is now roughly compatible with the scorecard worker API, and should be able to work within the scorecard infrastructure.

There is work to be done on config, testing the infra in docker compose, etc.


cmd/collect_signals/worker.go
```diff
@@ -0,0 +1,155 @@
...
+	s, err := scorer.FromConfig(scorer.NameFromFilepath(scoringConfigFile), f)
+	if err != nil {
+		return nil, fmt.Errorf("from config: %w", err)
+	}
+	return s, nil
+}
+
+func NewWorker(ctx context.Context, logger *zap.Logger, scoringEnabled bool, scoringConfigFile, scoringColumn string, collectOpts []collector.Option) (*collectWorker, error) {
+	logger.Info("Initializing worker")
+	logger.Debug("Creating collector")
```
oliverchang, line 128: Is this debug log needed if we have an info log right before this?

cmd/collect_signals/main.go
```diff
@@ -15,226 +15,107 @@
...
+		gcpDatasetName = collector.DefaultGCPDatasetName
+	}
 
-	// Complete the validation of args
-	if flag.NArg() != 1 {
-		logger.Error("Must have an input file specified.")
-		os.Exit(2)
+	// Determine whether scoring is enabled or disabled.
+	// It supports various "truthy" and "fasley" values. It will default to
+	// enabled.
```
oliverchang, line 84: isn't the default disabled here according to line 86 ?

oliverchang (approved):

cmd/collect_signals/worker.go
calebbrown: Fixed.

cmd/collect_signals/main.go
calebbrown: No, the default (i.e. when `criticalityConfig["scoring"] == ""`) is set to true in the `switch` statement - part of the issue is that Go doesn't let me create a `bool` without a value (the zero value is false).

I could have used `var scoringEnabled bool` but that is the same as `scoringEnabled := false`

Either way, I've added a comment and changed it to `true` make the default behaviour clearer.

oliverchang (approved):

PR #231, opened 2022-11-08
Add marker file support to enumerate_github

calebbrown:
The marker file stores the transformed filename of the file specified by -out.

This is useful when `[[runid]]` is used. It allows a consuming job to easily locate the most recently completed enumeration run without needing to enumerate objects in a bucket.


cmd/enumerate_github/main.go
```diff
@@ -151,6 +154,26 @@ func searchWorker(s *githubsearch.Searcher, logger *zap.Logger, queries, results
 	}
 }
 
+func writeMarker(ctx context.Context, markerFile, outFile string) (err error) {
+	marker, e := cloudstorage.NewWriter(ctx, markerFile)
+	if e != nil {
+		return fmt.Errorf("open marker: %w", err)
```
oliverchang, line 160: should this be `e` instead of `err` ?

cmd/enumerate_github/main.go
```diff
@@ -151,6 +154,26 @@ func searchWorker(s *githubsearch.Searcher, logger *zap.Logger, queries, results
...
+		if e := marker.Close(); e != nil {
+			// Return the error using the named return value if it isn't
+			// already set.
+			if err == nil {
+				err = fmt.Errorf("closing marker: %w", err)
+			}
+		}
+	}()
+	if _, e := fmt.Fprintln(marker, outFile); e != nil {
+		err = fmt.Errorf("writing marker: %w", err)
```
oliverchang, line 172: should the second `err` here be `e` instead?

cmd/enumerate_github/main.go
```diff
@@ -151,6 +154,26 @@ func searchWorker(s *githubsearch.Searcher, logger *zap.Logger, queries, results
...
+	marker, e := cloudstorage.NewWriter(ctx, markerFile)
+	if e != nil {
+		return fmt.Errorf("open marker: %w", err)
+	}
+	defer func() {
+		if e := marker.Close(); e != nil {
+			// Return the error using the named return value if it isn't
+			// already set.
+			if err == nil {
+				err = fmt.Errorf("closing marker: %w", err)
```
oliverchang, line 167: should the second `err` here be `e` instead? ****

cmd/enumerate_github/main.go
calebbrown: Fixed

cmd/enumerate_github/main.go
calebbrown: Fixed

cmd/enumerate_github/main.go
calebbrown: Fixed

oliverchang (approved):

PR #245, opened 2022-11-17
Support for CSV dumps from the production worker

calebbrown:
This PR adds support for CSV output alongside the JSON data being collected and inserted into BigQuery.

The CSV output is required because BigQuery doesn't cleanly support exports of tables with nested records.

Once the CSV is collected into shard files, a `csv_transfer` command will aggregate all the shards into a single CSV file and place it in a public bucket.

The following changes have been made to support this:

- a parallel output of CSV data alongside JSON data in `collect_signals`.
- re-use of "raw-result-bucket-url" config key to support this behaviour as an interim stop-gap.
- a `csv_transfer` tool introduced to aggregate and publish CSV data collected.
- small refactor to log init code to make it more reusable in production environments.


calebbrown:
FYI - this won't be merged until the current run has completed.

oliverchang (approved):

PR #248, opened 2022-11-22
Unit tests for Weighted Arithmetic Mean and Distribution

nathannaveen:
- Added unit tests for WAM.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -17,7 +17,6 @@ linters:
   enable:
     - asciicheck
     #- bodyclose  # Temporarily disabled.
-    - deadcode
     - depguard
     - dogsled
     #- errcheck  # Temporarily disabled.
@@ -59,7 +58,6 @@ linters:
     - unconvert
     - unparam
     - unused
-    - varcheck
     - whitespace
       #- wrapcheck
 linters-settings:
--- a/cmd/collect_signals/Dockerfile
+++ b/cmd/collect_signals/Dockerfile
@@ -22,7 +22,7 @@ COPY . ./
 FROM base AS collect_signals
 ARG TARGETOS
 ARG TARGETARCH
-RUN CGO_ENABLED=0 go build ./cmd/collect_signals
+RUN CGO_ENABLED=0 go build -buildvcs ./cmd/collect_signals
 RUN chmod -R 0775 /src/config/scorer/*
 
 FROM gcr.io/distroless/base:nonroot@sha256:533c15ef2acb1d3b1cd4e58d8aa2740900cae8f579243a53c53a6e28bcac0684
--- a/cmd/collect_signals/main.go
+++ b/cmd/collect_signals/main.go
@@ -91,6 +91,11 @@ func main() {
 	}
 	defer logger.Sync()
 
+	// Embed the commitID with all log messages.
+	if commitID != "" {
+		logger = logger.With(zap.String("commit_id", commitID))
+	}
+
 	// Extract the GCP project ID.
 	gcpProjectID, err := config.GetProjectID()
 	if err != nil {
--- a/cmd/collect_signals/vcs.go
+++ b/cmd/collect_signals/vcs.go
@@ -1,0 +1,40 @@
+// Copyright 2022 Criticality Score Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"runtime/debug"
+)
+
+const commitIDKey = "vcs.revision"
+
+var commitID = getCommitID()
+
+// getCommitID returns the vcs commit ID embedded in the binary when the
+// -buildvcs flag is set while building.
+func getCommitID() string {
+	info, ok := debug.ReadBuildInfo()
+	if !ok {
+		return ""
+	}
+
+	for _, setting := range info.Settings {
+		if setting.Key == commitIDKey {
+			return setting.Value
+		}
+	}
+
+	return ""
+}
--- a/cmd/collect_signals/worker.go
+++ b/cmd/collect_signals/worker.go
@@ -21,7 +21,10 @@ import (
 	"github.com/ossf/criticality_score/internal/signalio"
 )
 
-const collectionDateColumnName = "collection_date"
+const (
+	collectionDateColumnName = "collection_date"
+	commitIDColumnName       = "worker_commit_id"
+)
 
 type collectWorker struct {
 	logger          *zap.Logger
@@ -51,6 +54,9 @@ func (w *collectWorker) Process(ctx context.Context, req *data.ScorecardBatchReq
 		extras = append(extras, w.scoreColumnName)
 	}
 	extras = append(extras, collectionDateColumnName)
+	if commitID != "" {
+		extras = append(extras, commitIDColumnName)
+	}
 
 	var jsonOutput bytes.Buffer
 	jsonOut := signalio.JSONWriter(&jsonOutput)
@@ -101,6 +107,15 @@ func (w *collectWorker) Process(ctx context.Context, req *data.ScorecardBatchReq
 			Key:   collectionDateColumnName,
 			Value: jobTime,
 		})
+
+		// Ensure the commit ID is included with each record for helping
+		// identify which Git commit is associated with this record.
+		if commitID != "" {
+			extras = append(extras, signalio.Field{
+				Key:   commitIDColumnName,
+				Value: commitID,
+			})
+		}
 
 		// Write the signals to storage.
 		if err := jsonOut.WriteSignals(ss, extras...); err != nil {
--- a/cmd/criticality_score/README.md
+++ b/cmd/criticality_score/README.md
@@ -64,10 +64,10 @@ $ export GITHUB_TOKEN=ghp_abc,ghp_123
 BigQuery access requires the "BigQuery User" (`roles/bigquery.user`) role added
 to the account used, or be an "Owner".
 
-##### Option 1: `gcloud login`
+##### Option 1: `gcloud auth login`
 
-This option is useful during development. Run `gcloud login --update-adc` to
-login to GCP and prepare application default credentials.
+This option is useful during development. Run `gcloud auth login --update-adc`
+to login to GCP and prepare application default credentials.
 
 ##### Option 2: GCE Service Worker
 
--- a/infra/cloudbuild/release.yaml
+++ b/infra/cloudbuild/release.yaml
@@ -22,8 +22,10 @@ steps:
 - name: 'gcr.io/cloud-builders/docker'
   args: ['pull', 'gcr.io/openssf/criticality-score-csv-transfer:$COMMIT_SHA']
 
-# Start a Cloud Deploy replease
+# Start a Cloud Deploy replease. Ignore failures as this will get run daily,
+# even if latest commit hasn't changed.
 - name: gcr.io/google.com/cloudsdktool/cloud-sdk
+  allowFailure: true
   entrypoint: gcloud
   args:
     [
--- a/infra/envs/base/bq_transfer.yaml
+++ b/infra/envs/base/bq_transfer.yaml
@@ -17,8 +17,8 @@ kind: CronJob
 metadata:
   name: criticality-score-bq-transfer
 spec:
-  # At 02:00UTC on the 28th day of the month.
-  schedule: "0 2 28 * *"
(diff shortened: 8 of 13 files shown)


internal/scorer/algorithm/wam/wam_test.go
```diff
@@ -0,0 +1,130 @@
...
+				inputs: []*algorithm.Input{
+					{Weight: 0, Distribution: algorithm.LookupDistribution("linear"),
+						Source: algorithm.Value(algorithm.Field("0"))},
+					{Weight: 1, Distribution: algorithm.LookupDistribution("linear"),
+						Source: algorithm.Value(algorithm.Field("1"))},
+				},
+			},
+			want: 1,
+			args: args{
+				record: map[string]float64{"1": 1},
```
calebbrown, line 70: This should have a field for "0" so that it is clear the weight is excluding the value.

calebbrown (changes requested):
Thanks for implementing these tests!

internal/scorer/algorithm/wam/wam.go
```diff
@@ -15,6 +15,7 @@
 // The package wam implements the Weighted Arithmetic Mean, which forms the
 // basis of Rob Pike's criticality score algorithm as documented in
 // Quantifying_criticality_algorithm.pdf.
+
```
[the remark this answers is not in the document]
nathannaveen: Removed!

internal/scorer/algorithm/distribution_test.go
```diff
@@ -0,0 +1,84 @@
...
+// limitations under the License.
+
+package algorithm
+
+import (
+	"math"
+	"testing"
+)
+
+func TestLookupDistribution(t *testing.T) {
```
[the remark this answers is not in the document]
nathannaveen: Fixed

internal/scorer/algorithm/wam/wam_test.go
```diff
@@ -0,0 +1,130 @@
...
+	type fields struct {
+		inputs []*algorithm.Input
+	}
+	type args struct {
+		record map[string]float64
+	}
+	tests := []struct {
+		name   string
+		fields fields
+		args   args
```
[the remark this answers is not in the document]
nathannaveen: I took care of this.

internal/scorer/algorithm/wam/wam_test.go
```diff
@@ -0,0 +1,130 @@
...
+						Source: algorithm.Value(algorithm.Field("2"))},
+					{Weight: 3, Distribution: algorithm.LookupDistribution("linear"),
+						Source: algorithm.Value(algorithm.Field("3"))},
+					{Weight: 4, Distribution: algorithm.LookupDistribution("linear"),
+						Source: algorithm.Value(algorithm.Field("4"))},
+					{Weight: 5, Distribution: algorithm.LookupDistribution("linear"),
+						Source: algorithm.Value(algorithm.Field("5"))},
+				},
+			},
+			want: 3.6666666666666665,
```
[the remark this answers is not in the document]
nathannaveen: The code itself returns 3.66... so that is why I have the test wanting 3.66. What are your thoughts?

internal/scorer/algorithm/wam/wam_test.go
nathannaveen: I have changed that only to have a field of 0.

internal/scorer/algorithm/wam/wam_test.go
```diff
@@ -0,0 +1,130 @@
...
+						Source: algorithm.Value(algorithm.Field("0"))},
+				},
+			},
+			want: math.NaN(),
+			args: args{
+				record: map[string]float64{"1": 1},
+			},
+		},
+	}
+	for _, tt := range tests {
```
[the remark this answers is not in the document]
nathannaveen: 👍

internal/scorer/algorithm/distribution_test.go
```diff
@@ -0,0 +1,159 @@
...
+			distributionName:  "zipfian",
+			distributionValue: math.MaxInt64,
+
+			want: result{
+				name:  "zipfian",
+				value: 43.668272,
+			},
+		},
+	}
+	for _, tt := range tests {
```
calebbrown, line 129: Please use `test` instead of `tt`

internal/scorer/algorithm/distribution_test.go
```diff
@@ -0,0 +1,159 @@
...
+			}
+
+			normalizedValue := got.Normalize(tt.distributionValue)
+
+			// round to 6 decimal places for comparison
+			normalizedValue = math.Round(normalizedValue*1000000) / 1000000
+			wantValue := math.Round(tt.want.value*1000000) / 1000000
+
+			if got.String() != tt.want.name {
+				t.Errorf("The names are not equal LookupDistribution name = %s, want name %s",
```
calebbrown, line 144: nit: the text "The names are not equal" is not necessary, the rest of the string and the presence of the error makes the inequality obvious.

internal/scorer/algorithm/distribution_test.go
```diff
@@ -0,0 +1,159 @@
...
+					got.String(), tt.want.name)
+			}
+
+			if math.IsNaN(normalizedValue) && math.IsNaN(tt.want.value) {
+				// both are NaN, and we can't compare two NaNs together
+				return
+			}
+
+			if normalizedValue != wantValue {
+				t.Errorf("The values are not equal LookupDistribution value %f, want value %f",
```
calebbrown, line 154: nit: as above, the text "The values are not equal" is not necessary.

calebbrown (changes requested):

internal/scorer/algorithm/wam/wam_test.go
```diff
@@ -0,0 +1,126 @@
...
+				{
+					Weight: 4, Distribution: algorithm.LookupDistribution("linear"),
+					Source: algorithm.Value(algorithm.Field("4")),
+				},
+				{
+					Weight: 5, Distribution: algorithm.LookupDistribution("linear"),
+					Source: algorithm.Value(algorithm.Field("5")),
+				},
+			},
+			want:   3.6666666666666665,
```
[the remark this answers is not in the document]
nathannaveen: I realized why I got `3.666` and you got `1`, I ran the tests with `record: map[string]float64{"1": 1, "2": 2, "3": 3, "4": 4, "5": 5}`, but forgot to commit that so it was only `record: map[string]float64{"1": 1},`. I have fixed this, so the record is `record: map[string]float64{"1": 1, "2": 2, "3": 3, "4": 4, "5": 5}`.

internal/scorer/algorithm/wam/wam_test.go
```diff
@@ -0,0 +1,126 @@
...
+			record: map[string]float64{"1": 1},
+		},
+		{
+			name: "With zero weight",
+			inputs: []*algorithm.Input{
+				{
+					Weight: 0, Distribution: algorithm.LookupDistribution("linear"),
```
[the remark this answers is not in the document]
nathannaveen: I updated `LoadConfig()` from `config.go`, and `New()` from `wam.go` with the check for the weight being greater than `0`.

internal/scorer/algorithm/wam/wam_test.go
```diff
@@ -0,0 +1,126 @@
...
+			},
+			want:   1,
+			record: map[string]float64{"1": 1},
+		},
+		{
+			name: "with a single negative weight",
+			inputs: []*algorithm.Input{
+				{
+					Weight: -300, Distribution: algorithm.LookupDistribution("linear"),
+					Source: algorithm.Value(algorithm.Field("-300")),
```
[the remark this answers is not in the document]
nathannaveen: I fixed this by updating the negative weights.

internal/scorer/algorithm/distribution_test.go
```diff
@@ -0,0 +1,159 @@
...
+
+			if got == nil && tt.wantNil {
+				return
+			}
+
+			normalizedValue := got.Normalize(tt.distributionValue)
+
+			// round to 6 decimal places for comparison
+			normalizedValue = math.Round(normalizedValue*1000000) / 1000000
+			wantValue := math.Round(tt.want.value*1000000) / 1000000
```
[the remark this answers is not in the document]
nathannaveen: Agreed! I have taken care of this!

internal/scorer/algorithm/distribution_test.go
nathannaveen: Done!

internal/scorer/algorithm/distribution_test.go
nathannaveen: 👍

internal/scorer/algorithm/distribution_test.go
nathannaveen: 👍

internal/scorer/algorithm/wam/wam.go
```diff
@@ -18,16 +18,26 @@
...
 
 // New returns a new instance of the Weighted Arithmetic Mean algorithm, which
 // is used by the Pike algorithm.
 func New(inputs []*algorithm.Input) (algorithm.Algorithm, error) {
+	for _, i := range inputs {
+		if i.Weight <= 0 {
+			return nil, ErrWeightLessThanOrEqualToZero
+		}
+	}
+
```
calebbrown, line 40: This check belongs better in `config.go` or a new function for constructing Input instances.

internal/scorer/algorithm/wam/wam_test.go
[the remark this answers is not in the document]
calebbrown: Thanks for fixing. That makes more sense.

internal/scorer/algorithm/wam/wam_test.go
```diff
@@ -0,0 +1,115 @@
...
+				},
+				{
+					Weight: 5, Distribution: algorithm.LookupDistribution("linear"),
+					Source: algorithm.Value(algorithm.Field("5")),
+				},
+			},
+			want:   3.6666666666666665,
+			record: map[string]float64{"1": 1, "2": 2, "3": 3, "4": 4, "5": 5},
+		},
+		{
```
calebbrown, line 60: If these are blocked when the Input is created then I don't think these "negative weight" or "zero weight" tests need to be here.

However there should be tests here for edge cases, for example:

- no fields matching the record
- partial matches between the defined fields and the record

calebbrown (changes requested):

internal/scorer/algorithm/wam/wam.go
nathannaveen: I don't understand what you mean, could you please abbreviate?

calebbrown: Sure.

Currently, this check will only be applied to the Weighted Arithmetic Mean.

The check should be applied across all possible algorithms as zero weights are no-ops and negative weights are confusing and error prone.

The most sensible place to have this check is inside the config.go in `UnmarshalYAML()` while the Input is being parsed.

However, another place it could be added is in `algorithm/input.go` by creating a 'NewInput()` method that does the validation which would be called from `ToAlgorithmInput()` in config.go.

internal/scorer/config.go
```diff
@@ -121,6 +123,13 @@ func LoadConfig(r io.Reader) (*Config, error) {
 	if err := yaml.Unmarshal(data, c); err != nil {
 		return nil, err
 	}
+
+	for _, i := range c.Inputs {
+		if i.Weight <= 0 {
+			return nil, wam.ErrWeightLessThanOrEqualToZero
+		}
```
calebbrown, line 130: See my comment re:this error check.

Additionally, the error `ErrWeightLessThanOrEqualToZero` should not be pulled from `wam`.

calebbrown (changes requested):

internal/scorer/algorithm/wam/wam.go
nathannaveen: Thank you for the explanation. I have added the check into `UnmarshalYAML()`.

internal/scorer/algorithm/wam/wam_test.go
nathannaveen: I have removed the negative weight and zero weight tests, and have put in the tests `"fields not matching the record"` and `"some fields matching the record"`.

internal/scorer/config.go
nathannaveen: 👍

internal/scorer/config.go
```diff
@@ -52,6 +52,11 @@ func (i *Input) UnmarshalYAML(value *yaml.Node) error {
 		return errors.New("field must be set")
 	}
 	*i = Input(*raw)
+
+	if i.Weight <= 0 {
+		return errors.New("weight must be greater than 0")
+	}
+
```
calebbrown, line 59: nit: please but this under the `raw.Field` check and use `raw.Weight` instead of `i.Weight`

Otherwise this is exactly what I had in mind! thanks :)

calebbrown (changes requested):
Looking good, just one minor change!

internal/scorer/config.go
nathannaveen: Took care of it, thanks

calebbrown (approved):
Thanks!

calebbrown (approved):

nathannaveen:
> Thanks!

Yay! Thanks for all your reviews! Merged! 🥳

PR #250, opened 2022-11-24
Build a more formal deploy process and improve infra config

calebbrown:
Create/improve the deploy process and improve the kubernetes infrastructure config.

- Kustomize is used to avoid needing to re-define k8s configs for each environment.
- Cloud Deploy configs have been added to support using Cloud Deploy for managing the release process.
  - Two targets/profiles have been added "staging" and "prod".
  - "staging" is used first, and a working release is promoted to "prod".
- A new cloudbuild config for triggering releases has been added. This will be run on a fixed schedule once a day. It should abort if the release already exists for the given commit ID.
- Image names and tags are injected into the k8s configs.
  - Scorecard images are hardcoded to the same commit ID as the go.mod
  - Criticality Score images are set during the Cloud Deploy release.
- To ensure Scorecard image tags match go.mod a check has been added that will run in a GitHub CI Workflow


oliverchang (approved):
very nice!

PR #257, opened 2022-11-25
Unit tests - internal/githubapi/errors

nathannaveen:
- Unit tests for errors
- Refactored errors


internal/githubapi/errors.go
```diff
@@ -95,11 +95,7 @@ func (e *GraphQLErrors) Errors() []GraphQLError {
 
 // Is implements the errors.Is interface.
 func (e *GraphQLErrors) Is(target error) bool {
-	if target == e {
```
nathannaveen, line 98: `target` will never equal `e` because `target == error` and `e == *GraphQLErrors`

internal/githubapi/errors.go
```diff
@@ -95,11 +95,7 @@ func (e *GraphQLErrors) Errors() []GraphQLError {
...
-		// Identity test is always true.
-		return true
-	}
-	if target == ErrGraphQLNotFound {
```
nathannaveen, line 102: The correct way to compare errors is `errors.Is()`

internal/githubapi/errors_test.go
```diff
@@ -0,0 +1,184 @@
...
+		name string
+		err  error
+		want int
+	}{
+		{
+			name: "nil error",
+			want: 0,
+		},
+		{
+			name: "non-nil error that is not in ErrorResponse",
```
calebbrown, line 36: nit: "... is not *an* ErrorResponse"

internal/githubapi/errors_test.go
```diff
@@ -0,0 +1,184 @@
...
+			},
+			want: "3 GraphQL errors",
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			defer func() {
+				r := recover()
+				if r != nil && !test.wantPanic {
+					t.Fatalf("Error() panic: %v, %s", r, test.name)
```
calebbrown, line 93: test.name is not required here.

The message should read "Error() panicked: %v, want no panic"

internal/githubapi/errors_test.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+}
+
+func TestGraphQLErrors_HasType(t *testing.T) {
+	tests := []struct { //nolint:govet
+		name   string
+		errors []GraphQLError
+		t      string
+		want   bool
+	}{
```
calebbrown, line 112: Please add two more tests:

1. two GraphQLErrors one with NOT_FOUND, one without.
2. two GraphQLErrors where their Type fields are set to something other than the search.

internal/githubapi/errors.go
calebbrown: That is true generally, however, in this case we are implementing the `Is(error) errror` interface that makes `errors.Is` work.

Comparing directly using `target == ErrGraphQLNotFound` is the right way to do this.

The docs on https://pkg.go.dev/errors#Is say `"An Is method should only shallowly compare err and the target and not call Unwrap on either."` - calling `errors.Is()` in this context would cause an `Unwrap` to occur violating this.

The docs even show a similar example: `target == fs.ErrExist`.

internal/githubapi/errors.go
calebbrown: Good point.

calebbrown (changes requested):

internal/githubapi/errors.go
nathannaveen: Agreed, thanks.

internal/githubapi/errors_test.go
nathannaveen: I have changed the error message.

internal/githubapi/errors_test.go
nathannaveen: I have added the two tests.

internal/githubapi/errors_test.go
nathannaveen: 👍

calebbrown (changes requested):

internal/githubapi/errors_test.go
```diff
@@ -0,0 +1,200 @@
...
+				errors: test.errors,
+			}
+			if got := e.HasType(test.t); got != test.want {
+				t.Errorf("HasType() = %v, want %v", got, test.want)
+			}
+		})
+	}
+}
+
+func TestGraphQLErrors_Is(t *testing.T) {
```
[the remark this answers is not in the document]
nathannaveen: I have added the test case. Thank you

calebbrown (approved):

PR #266, opened 2022-11-30
Makefile: Included gofumpt target

nathannaveen:
- Included gofumpt target to format the code.
- Will make it easier for new contributors.


Makefile
```diff
@@ -53,3 +53,6 @@ build/docker/csv-transfer:
 install/deps:  ## Installs all dependencies during development and building
 	@echo Installing tools from tools/tools.go
 	@cd tools; cat tools.go | grep _ | awk -F'"' '{print $$2}' | xargs -tI % go install %
+gofumpt: ## Run gofumpt on all go files
+	@which gofumpt || (echo "gofumpt is not installed. Please install it using 'go install mvdan.cc/gofumpt@latest'" && exit 1)
+	gofumpt -w -l .
```
calebbrown, line 58: This is okay, but it would be better if it followed the same pattern as `lint`.

1. in `/tools/` add mvdan.cc/gofumpt go the source file and make sure you run `go get` inside the dir to add it to the go.mod/go.sum files.
1. in `Makefile` have a var link `GOLANGCI_LINT` for `GOFUMPT`
1. duplicate the `lint` block, but call it `format` instead of `gofumpt`.

Doing this will use the `install/deps` target to ensure `gofumpt` is available.

Also can you place it after the `lint` section, before the `build/docker` section.

Thanks!

calebbrown (changes requested):

Makefile
nathannaveen: I have updated the `gofumpt`. Thanks!

Makefile
```diff
@@ -37,6 +38,12 @@ lint:  ## Run linter
 lint: $(GOLANGCI_LINT)
 	$(GOLANGCI_LINT) run -c .golangci.yml
 
+.PHONY: gofumpt
+$(GOFUMPT): install/deps
+gofumpt:  ## Run formatter
+gofumpt: $(GOFUMPT)
+	$(GOFUMPT) -w -l .
```
calebbrown, line 45: One minor change - can the name of the make target be `format:` instead of `gofumpt:` thanks!

calebbrown (changes requested):

Makefile
nathannaveen: Changed `gofumpt` to `format`. Thanks!

calebbrown (approved):

PR #275, opened 2022-12-06
Fixed Bug in registry/EmptySets

nathannaveen:
- The exists map is always empty so it can't contain anything.


calebbrown:
I think you've found a bug. At the end of the loop there should be:

```go
exists[c.EmptySet().Namespace()] = empty{}
```

This is important for when GitLab support is added and returns IssueSet and RepoSet.

nathannaveen:
> I think you've found a bug. At the end of the loop there should be:
> 
> ```go
> exists[c.EmptySet().Namespace()] = empty{}
> ```
> 
> This is important for when GitLab support is added and returns IssueSet and RepoSet.

Fixed the bug!

calebbrown (approved):
Thanks!

PR #286, opened 2022-12-16
Included tests for internal/collector/github/factory

nathannaveen:
- Included tests for internal/collector/github/factory
- Included mocks for query
- Added interface for query


nathannaveen:
I removed:

https://github.com/ossf/criticality_score/blob/95942e03caa6efc1616c47e8f1c446c9fb256914/internal/collector/github/repo.go#L45-L48

I removed this because `r.BasicData` is never going to be `nil`. The `New` function in the `factory.go` file is the only place `init` is called. `r.BasicData` is never set when the code calls `r.init()`because `r` is initialized without `BasicData`:

``` go
r := &repo{
    client:  f.client,
    origURL: u,
    logger:  f.logger.With(zap.String("url", u.String())),
}
```

nathannaveen:
@calebbrown A friendly ping!

PR #292, opened 2022-12-28
Included Build Targets for Binaries

nathannaveen:
- Included Make file targets for the binaries
- Included the check in github ci


nathannaveen:
@calebbrown A friendly ping!

PR #306, opened 2023-01-11
Included tests for internal/signalio/json.go

nathannaveen:
- Included tests for internal/signalio/json.go
- Removed un-necessary error check


nathannaveen:
I have removed:

https://github.com/ossf/criticality_score/blob/b148d93395c5cc004fcb2237a9b008b45d8c4937/internal/signalio/json.go#L54-L57

I have replaced it with: `nsData := d.(map[string]any)`.

I replaced this part because `d` is always set to a `map[string] any` before the check, so it will never be `!ok`.

internal/signalio/json_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+func (t testJSONWriterSet) Namespace() signal.Namespace {
+	return "test"
+}
+
+type mockWriterJSON struct{}
+
+func (m *mockWriterJSON) Write(p []byte) (n int, err error) {
+	return 0, nil
+}
```
calebbrown, line 38: Use a `bytes.Buffer` to avoid needing to implement the `io.Writer` interface.
It also allows you to inspect what was written.

internal/signalio/json_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+func Test_jsonWriter_WriteSignals(t *testing.T) {
+	type args struct {
+		signals []signal.Set
+		extra   []Field
+	}
+	test := struct {
```
calebbrown, line 45: May I suggest restructuring these tests into the following the test functions as it isn't really clear what this function is trying to test:

1. An end-to-end style test that:
  - creates the writer with a `bytes.Buffer`.
  - passes in test data for all the parameters.
  - checks the output matches a JSON string (e.g. `want := "{...}"`).
2. A test that checks a writer error with a `io.Writer` implementation that always returns an error (like `iotest.ErrReader`, but an equivalent for `io.Writer` doesn't exist).
3. (Optional) a test that exercises encoding errors.

These should ideally not contain the test struct building at the start as it isn't really necessary.

internal/signalio/json.go
```diff
@@ -51,10 +50,9 @@ func (w *jsonWriter) WriteSignals(signals []signal.Set, extra ...Field) error {
 				d = make(map[string]any)
 				data[ns] = d
 			}
-			nsData, ok := d.(map[string]any)
-			if !ok {
-				return fmt.Errorf("failed to get map for namespace: %s", ns)
-			}
+			nsData := d.(map[string]any)
+			// we don't need to check for an error here (the above line) because d is always a map[string]any
```
calebbrown, line 54: Might be worth extending the comment to say that "if this fails in the future the assignment to nsData below will fail"

calebbrown (changes requested):

PR #307, opened 2023-01-12
Updated tests for internal/signalio/type

nathannaveen:
- Changed the package name for the tests from signalio_test to signalio
- Updated test
- 100% coverage on `internal/signalio/type.go`


internal/signalio/type_test.go
```diff
@@ -102,3 +107,63 @@ func TestTypeUnmarshalText(t *testing.T) {
 		})
 	}
 }
+
+func TestWriterType_New(t *testing.T) {
```
calebbrown, line 111: I'm not sold on changing the package from `signalio_test` to `signalio` to access the internal types to implement this test.

Some alternative approaches:
- check the result of `New(...)` against `JSONWriter(...)`
- update the `signalio.Writer` interface to include a `Type() signalio.WriterType` method
- call `WriteSignalSet(...)` with a `bytes.Buffer` and compare the output

calebbrown (changes requested):

internal/signalio/type_test.go
nathannaveen: After taking a second look, I realized that replacing some struct calls would allow me to use the package `signalio_test` instead of `signalio`. Thanks!

calebbrown (approved):

PR #309, opened 2023-01-13
Included tests for internal/log/env

nathannaveen:
- Included tests for internal/log/env
- Included licenses for `env.go` and `config.go`


internal/log/env_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+func TestLookupEnv(t *testing.T) {
+	tests := []struct {
+		name string
+		text string
+		want Env
+	}{
+		{"dev", "dev", DevEnv},
+		{"gcp", "gcp", GCPEnv},
+		{"unknown", "unknown", UnknownEnv},
```
calebbrown, line 31: I suggest an empty string in this test as well.

internal/log/env_test.go
```diff
@@ -0,0 +1,110 @@
...
+		e    Env
+		want []byte
+	}{
+		{"dev", DevEnv, []byte("dev")},
+		{"unknown", UnknownEnv, []byte("unknown")},
+		{"empty", math.MaxInt, []byte("unknown")},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, _ := tt.e.MarshalText() // this function never returns an error
```
calebbrown, line 103: There should be a check that err is always `nil` so the assertion is checked.

internal/log/env_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+			if !reflect.DeepEqual(got, tt.want) {
```
calebbrown, line 105: `bytes.Equal()` a better choice than `DeepEqual`

internal/log/env_test.go
```diff
@@ -0,0 +1,110 @@
...
+			wantErr: true,
+		},
+		{
+			name:  "dev",
+			text:  []byte("dev"),
+			value: "dev",
+		},
+	}
+	for _, tt := range tests {
+		t.Setenv(string(tt.text), tt.value)
```
calebbrown, line 82: I don't think this call does anything and probably should be removed, along with `tt.value`

internal/log/env_test.go
```diff
@@ -0,0 +1,110 @@
...
+			value:   "unknown",
+			wantErr: true,
+		},
+		{
+			name:  "dev",
+			text:  []byte("dev"),
+			value: "dev",
+		},
+	}
+	for _, tt := range tests {
```
calebbrown, line 81: nit: `test` is a better than `tt` as `tt` is very similar to `t`

calebbrown (changes requested):

internal/log/env_test.go
nathannaveen: I have added The test `{"empty", "", UnknownEnv}`.

internal/log/env_test.go
nathannaveen: Thanks, I forgot about that when writing the test! I have added the error check `err != nil`.

internal/log/env_test.go
nathannaveen: I have changed it from `reflect.DeepEqual()` to `bytes.Equal()`.

internal/log/env_test.go
nathannaveen: I have removed the `t.Setenv()` but I kept the `tt.value` because I added a check `test.e.String() != test.value`.

internal/log/env_test.go
nathannaveen: I have changed all the `tt`'s to `test`.

calebbrown (approved):

PR #310, opened 2023-01-16
Included tests for internal/collector/source

nathannaveen:
- Included some tests for internal/collector/source


nathannaveen:
I couldn't complete the `Get()` tests because when mocking `bqAPI` it resulted in the files calling each other in a cycle. I also didn't complete the `NewSource()` tests because I can't mock `bigquery.NewClient()`.

nathannaveen:
@calebbrown A friendly ping!

internal/collector/depsdev/source_test.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+}
+
+func Test_depsDevSource_Get(t *testing.T) {
+	type args struct {
+		ctx       context.Context
+		rHostName string
+		jobID     string
+	}
+	tests := []struct { //nolint:govet
```
calebbrown, line 60: Just wondering if you intend to add more test cases here?

If not then the struct, the loop and the `t.Run` call are not necessary.

internal/collector/depsdev/source_test.go
```diff
@@ -0,0 +1,184 @@
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ctrl := gomock.NewController(t)
+			defer ctrl.Finish()
+
+			r := mocks.NewMockRepo(ctrl)
+
+			r.EXPECT().URL().Return(&url.URL{Host: tt.args.rHostName}).AnyTimes()
```
calebbrown, line 86: For an interface with a single method, I think the MockRepo is not needed.

All that is needed to satisfy the interface is:
```go
type fakeRepo string
func (t fakeRepo) URL() string { return t}
```

calebbrown (changes requested):

internal/collector/depsdev/source_test.go
```diff
@@ -0,0 +1,184 @@
...
+		{
+			name: "random",
+			u: &url.URL{
+				Host: "random",
+			},
+			wantProjectName: "",
+			wantProjectType: "",
+		},
+	}
+	for _, tt := range tests {
```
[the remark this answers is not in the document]
nathannaveen: I have changed all the `tt`'s to `test`.

internal/collector/depsdev/source_test.go
nathannaveen: I don't currently intend to make more tests for this function so I have removed the loop and `t.Run`.

internal/collector/depsdev/source_test.go
nathannaveen: You are right, I have removed the mock and included something similar to what you have shown.

PR #316, opened 2023-01-30
Removed GlobalRegistry from registry.go

nathannaveen:
Removed the `GlobalRegisty` variable from `internal/scorer/algorithm/registry.go` because it isn't good practice to use global variables. I was able to remove `GlobalRegistry` because the only places any function in `registry.go` was being used was in `wam.go` and `config.go`.

The `Register()` function (From `registry.go`) was only being used in the `init()` function (From `wam.go`). The `NewALgorithm()` function (From `registry.go`) was only being used in the `Algorithm()` function (From `config.go`). Since those were the only places any function from `registry.go` was being used externaly I created a `NewRegistry()` in `Algorithm()` and then registered it, and returned `r.NewAlgorithm()`.

Because of creating a `NewRegistry()` and registering it in `Algorithm` I removed the `init()` function in `wam.go` because it wasn't being used. I also removed the `Register()` and `NewAlgorithm()` functions from `registry.go` because they weren't being used. I also removed `GlobalRegisty`.

A key motivation for this decision was to eliminate the `init()` function, due to the considerations outlined in Peter Bourgon's blog post, "Theory of Modern Go" (https://peter.bourgon.org/blog/2017/06/09/theory-of-modern-go.html).

### DIFF
--- a/internal/scorer/algorithm/registry.go
+++ b/internal/scorer/algorithm/registry.go
@@ -16,9 +16,6 @@ package algorithm
 
 import "fmt"
 
-// GlobalRegistry is the global, application wide, registry for all algorithms.
-var GlobalRegistry = NewRegistry()
-
 // Registry is used to map a name to a Factory that creates Algorithm instances
 // for the given name.
 type Registry struct {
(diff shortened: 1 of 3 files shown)


calebbrown:
Thanks Nathan! This change generally looks good, just one minor nit.

nathannaveen:
Hi Caleb, thanks for the review. Just wanted to double check, did you add the nit to it?

calebbrown (changes requested):

calebbrown:
> Hi Caleb, thanks for the review. Just wanted to double check, did you add the nit to it?

🤦 woops, the "pending" tag is very easy to miss. Added now.

calebbrown (approved):

PR #264, opened 2022-11-28
Unit Tests - scorer

nathannaveen:
- Included unit tests for scorer
- Included test data for scorer
- Included an additional validation for scorer

Signed-off-by: nathannaveen <42319948+nathannaveen@users.noreply.github.com>

PR #317, opened 2023-01-31
Refactored internal/scorer/scorer

nathannaveen:
- Refactored `internal/scorer/scorer.go` to improve readablilty.


internal/scorer/scorer.go
```diff
@@ -103,10 +83,19 @@ func NameFromFilepath(filepath string) string {
...
 	f := path.Base(filepath)
 	ext := path.Ext(f)
-	// Strip the extension and convert to lowercase
-	f = strings.ToLower(strings.TrimSuffix(f, ext))
-	// Change any non-alphanumeric character into an underscore
-	f = regexp.MustCompile("[^a-z0-9_]").ReplaceAllString(f, "_")
+
+	modified := func(r rune) rune {
+		// Change any non-alphanumeric character into an underscore
+		if !unicode.IsDigit(r) && !unicode.IsLetter(r) {
```
calebbrown, line 89: This condition does not match `"[^a-z0-0_]"` as it will preserve unicode letters in the filename in the output.

For example a filename of "configuración_básica.yaml" previously would have resulted in `configuraci_n_b_sica_score`, now it produces `configuración_básica_score` (arguably better).

The milestone 1 doc doesn't specify the encoding of field names, but does permit Unicode strings.

Given that this changes the behaviour it should be covered by a change in tests too.

internal/scorer/scorer.go
```diff
@@ -103,10 +83,19 @@ func NameFromFilepath(filepath string) string {
 	// Get the name of the file used, without the path
 	f := path.Base(filepath)
 	ext := path.Ext(f)
-	// Strip the extension and convert to lowercase
-	f = strings.ToLower(strings.TrimSuffix(f, ext))
-	// Change any non-alphanumeric character into an underscore
-	f = regexp.MustCompile("[^a-z0-9_]").ReplaceAllString(f, "_")
+
+	modified := func(r rune) rune {
```
calebbrown, line 87: Consider using `modifier` for the function name.

calebbrown (changes requested):

internal/scorer/scorer.go
nathannaveen: Which tests do you mean? I don't see any tests covering `NameFromFilepath()`, and I can't see any tests on any place it is being used.

internal/scorer/scorer.go
nathannaveen: I have changed `modified` to `modifier`.

internal/scorer/scorer.go
```diff
@@ -50,31 +50,11 @@ func FromConfig(name string, r io.Reader) (*Scorer, error) {
...
-			case uint16:
-				record[k] = float64(r)
-			case uint32:
-				record[k] = float64(r)
-			case uint64:
-				record[k] = float64(r)
-			case byte:
-				record[k] = float64(r)
+			case float64, float32, int, int16, int32, int64, uint, uint16, uint32, uint64, byte:
+				record[k] = r.(float64)
```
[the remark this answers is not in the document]
nathannaveen: You are right, it doesn't work, so I decided to try something new, using [`reflect.Value.Convert`](https://pkg.go.dev/reflect#Value.Convert). I have also tested it out, in a Playground, https://go.dev/play/p/ArBOmwGLUlh.

internal/scorer/scorer.go
nathannaveen: Oh, I misread your comment. I will write tests for it!

nathannaveen:
I have included tests for `scorer`, but have closed #264 because it is similar and would overlap.

internal/scorer/scorer.go
nathannaveen: I have written the tests.

internal/scorer/scorer.go
```diff
@@ -50,32 +56,10 @@ func FromConfig(name string, r io.Reader) (*Scorer, error) {
...
-				record[k] = float64(r)
-			case uint32:
-				record[k] = float64(r)
-			case uint64:
-				record[k] = float64(r)
-			case byte:
-				record[k] = float64(r)
-			}
+			// This takes the value of v and converts it to a float64.
+			record[k] = reflect.ValueOf(v).Convert(reflect.TypeOf(float64(0))).Float()
```
calebbrown, line 62: Unfortunately, this way of converting `v` to a `float64` has a subtle bug where types such as `time.Time` and `string` will cause the scoring to crash.


As it stands, I'm not sure I prefer this over the original `switch`.

The list of repeated type conversions doesn't look pretty, but it is functional and easy to understand.

calebbrown (changes requested):

nathannaveen:
@calebbrown A friendly ping!

calebbrown (approved):

Issue #288, opened 2022-12-21
Can only install the standalone CLI from @main


pnacht:
The README suggests installing the CLI with

```shell
go install github.com/ossf/criticality_score/cmd/criticality_score
```

However, I get an error:

```shell
go: 'go install' requires a version when current directory is not in a module
Try 'go install github.com/ossf/criticality_score/cmd/criticality_score@latest' to install the latest version
```

However, when I then try `@latest` (and `@v1.0.7`, the latest release tag), I get another error:

```shell
go: github.com/ossf/criticality_score/cmd/criticality_score@latest: module github.com/ossf/criticality_score@latest found (v1.0.7), but does not contain package github.com/ossf/criticality_score/cmd/criticality_score
```

The only method I've found that works is using `@main`, but that's not an optimal solution since it requires that the main branch be in a usable state.

calebbrown:
Thanks for the feedback @pnacht - I am hoping to set up a binary release process in the not-to-distant future which should make this problem go away.

The "1.0.7" version is a legacy reference to the old Python code and needs to go away.

noah-de:
Is there any way that I could contribute to this effort?

gabibguti:
I believe we can close this issue for now.

calebbrown:
Done, thanks!

PR #319, opened 2023-02-01
New user 'quickstart' instructions

noah-de:
As a new user, I wanted to make use of this tool and understand how it works.
My experience was similar to what is discussed in issue https://github.com/ossf/criticality_score/issues/288 https://github.com/ossf/criticality_score/issues/288
While this doesn't exactly fix the codebase or provide a new release process, it does provide a helpful message on how to quickly get started with what exists today.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,13 +45,15 @@ open source project:
 - We are looking for community ideas to improve upon these parameters.
 - There will always be exceptions to the individual reasoning rules.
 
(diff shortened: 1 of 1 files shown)


noah-de:
https://github.com/ossf/criticality_score/pull/319 doesn't seem necessary because https://github.com/ossf/criticality_score/pull/320 was merged

PR #327, opened 2023-02-14
Included Wrapcheck linter

nathannaveen:
Currently, a lot of errors that are returned are just returning the error and not giving any description to what the error is when debugging (i.e. `return err`). A solution to this is to use the golang linter for wrapcheck.

- Included the wrapcheck linter
- Updated settings to ignore previous wrapcheck issues

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -11,7 +11,9 @@ issues:
   # Set to 0 to disable.
   # Default: 3
   max-same-issues: 0
-  new-from-rev: ""
+  # Exclude previously existing issues from the wrapcheck report
+  new: true
+  new-from-rev: HEAD
 linters:
   disable-all: true
   enable:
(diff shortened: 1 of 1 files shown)


.golangci.yml
```diff
@@ -11,7 +11,9 @@ issues:
   # Set to 0 to disable.
   # Default: 3
   max-same-issues: 0
-  new-from-rev: ""
+  # Exclude previously existing issues from the wrapcheck report
```
calebbrown, line 14: nit: this comment should just say "Exclude previously existing issues from the report", as it applies to all reports (although in this case only the `wrapcheck` has violations)

nathannaveen: Took care of it. Thanks!

calebbrown (approved):

PR #331, opened 2023-02-23
Setup goreleaser to generate Go binaries.

calebbrown:
Introduces goreleaser to automate the creation of binary artifacts when a new release tag is created.

Additionally uses cosign to automatically sign releases.

### DIFF
--- a/.github/workflows/binary-release.yml
+++ b/.github/workflows/binary-release.yml
@@ -6,12 +6,16 @@ on:
       - 'v*'
 
 permissions:
-  contents: write  # needed for goreleaser to create the release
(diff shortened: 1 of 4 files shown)


oliverchang (approved):
LGTM, but maybe also take a look at https://github.com/google/osv-scanner/blob/main/.github/workflows/goreleaser.yml 

to see if there's any things that can be simplified there by using  `slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml`

calebbrown:
PTAL - I switched over to the SLSA generator

oliverchang (approved):
nice!

Issue #343, opened 2023-03-03
Most obvious CSV data is two years out of date


nedbat:
The home page says:

> CSV data
> The data is available on Google Cloud Storage and can be downloaded via:
> 
> web browser: [commondatastorage.googleapis.com/ossf-criticality-score/index.html](https://commondatastorage.googleapis.com/ossf-criticality-score/index.html)

That page has handy per-language files, but they are dated 2020-12-30.  Newer data should be made easier to find, or at least stale data should be removed as an attractive nuisance.

calebbrown:
I've rearranged the objects in the bucket - does this help?

nedbat:
Are the files in "archive" the same 2020 files?  It helps in that the old files are now in "archive", but now their dates are 2023, which is itself misleading.  Is there a reason to keep the old files at all?  Why not produce "top 200" files for current data?

calebbrown:
I've put them in a folder roughly correlating to the date they were originally created.

As for producing "top 200" files for current data - I'm interested in how you might be using these.

I had been leaning towards not producing top-200 sets for each language group, and just supplying a script for producing them locally.

However if the top-200 sets are providing value, I'm more than happy to work on getting these produced automatically.

nedbat:
TBH, I'm new to this data set, and am not sure how I would use the data.  I wrote this issue as some feedback from a new user trying to understand the data set.  The link from the README sounds enticing, then I am looking at a raw web server page with old files.  My suggestion is simply to present the data *you* value in a way that makes it easy for people to find it and understand it.

PR #344, opened 2023-03-03
Updated link for CSV data

nathannaveen:
- Fixes https://github.com/ossf/criticality_score/issues/343


calebbrown:
Thanks for suggesting this, but I'm reluctant to merge this change.

Hopefully in the next 2-4 weeks, results will be continuously generated. So the link could be out-of-date.

PR #356, opened 2023-03-20
Included Comments for Wam

nathannaveen:
- Included that "Weighted Arithmetic Mean" is also called "Weighted Average" for new users.
- Edited name scheme inside the function Score because `totalWeight` should be the name of the numerator in the return value.


internal/scorer/algorithm/wam/wam.go
```diff
@@ -23,26 +23,28 @@ import (
 
 const Name = "weighted_arithmetic_mean"
 
-type WeighetedArithmeticMean struct {
+// WeightedArithmeticMean is an implementation of the Weighted Arithmetic.
+// "Weighted Arithmetic Mean" is also known as "Weighted Average".
```
calebbrown, line 27: The first sentence might need the work "Mean" at the end.

A blank line between each sentence would be clearer.

Also - a link to the Wikipedia page may be useful too!

internal/scorer/algorithm/wam/wam.go
```diff
@@ -23,26 +23,28 @@ import (
...
+type WeightedArithmeticMean struct {
```
calebbrown, line 28: lol. nice catch on the typo.

internal/scorer/algorithm/wam/wam.go
```diff
@@ -23,26 +23,28 @@ import (
...
+func (p *WeightedArithmeticMean) Score(record map[string]float64) float64 {
+	var totalCount float64
 	var totalWeight float64
-	var s float64
 	for _, i := range p.inputs {
 		if v, ok := i.Value(record); ok {
-			totalWeight += i.Weight
-			s += i.Weight * v
+			totalCount += i.Weight
+			totalWeight += i.Weight * v
```
calebbrown, line 46: I think these names are back to front. I can appreciate that it makes the division at the end look more like an average but the use of `totalWeight` is confusing.

So maybe:

```go
totalWeight += i.Weight
weightedInputSum += i.Weight * v
```

Or maybe this is better
```
itemCount += i.Weight
itemSum += i.Weight * v
```

internal/scorer/algorithm/wam/wam.go
nathannaveen: Thanks for the "Mean" catch!

Included a blank line and the Wikipedia page link.

internal/scorer/algorithm/wam/wam.go
nathannaveen: 👍

calebbrown (approved):

PR #393, opened 2023-05-11
Reduce resources requested by the worker.

calebbrown:


infra/envs/base/worker.yaml
```diff
@@ -33,10 +33,10 @@ spec:
           imagePullPolicy: Always
           resources:
             requests:
-              memory: 5Gi
+              memory: 2Gi
```
spencerschrock, line 36: fyi: Azeem took a different approach in https://github.com/ossf/scorecard/pull/2978. 

Does it make sense to remove the requests to allow GKE to handle scaling? You're of course more familiar with your resource needs, so feel free to resolve conversation as needed.

spencerschrock (approved):

infra/envs/base/worker.yaml
calebbrown: Sounds good. I removed the memory limits.

spencerschrock (approved):

PR #399, opened 2023-05-17
Increase the number of stars for the enumeration to reduce the number of repos.

calebbrown:
Brings the number of repos in enumeration to just under 400k


bunos06 (approved):

spencerschrock (approved):
LGTM other than commit message. Should be increase number of stars right?

PR #505, opened 2023-12-18
Add a local, single worker implementation of a work loop.

calebbrown:
This implementation allows the collect signals implementation to be run without needing GCP PubSub.

Repo URLs will be retrieved using an initContainer to execute a command to fetch them from the GCS bucket.


cmd/collect_signals/localworker/loop.go
```diff
@@ -0,0 +1,232 @@
...
+	criticalityConfig, err := config.GetCriticalityValues()
+	if err != nil {
+		return nil, fmt.Errorf("criticality score config: %w", err)
+	}
+
+	inFile := criticalityConfig[configLocalURLDataFile]
+	if inFile == "" {
+		return nil, fmt.Errorf("%q config not set", configLocalURLDataFile)
+	}
+	f, err := os.Open(inFile)
```
oliverchang, line 105: Who closes this?

oliverchang (approved):

cmd/collect_signals/localworker/loop.go
calebbrown: Oh thanks! I forgot to close the iterator which closes the file.

cmd/collect_signals/localworker/loop.go
```diff
@@ -0,0 +1,238 @@
...
+}
+
+func (l *WorkLoop) process(ctx context.Context, req *data.ScorecardBatchRequest, bucketURL string) error {
+	exists, err := resultExists(ctx, req, bucketURL)
+	if err != nil {
+		return fmt.Errorf("result exists check: %w", err)
+	}
+
+	// Sanity check - make sure we are not re-processing an already processed request.
+	if exists {
```
maxfisher-g, line 137: should this be logged?

cmd/collect_signals/main.go
```diff
@@ -39,6 +41,58 @@ const (
...
+// getWorkLoop chooses between either scorecard's worker.WorkLoop implementation
+// or the internal localworker.WorkLoop implementation. The localworker.WorkLoop
+// is chosen if the "local" criticality config variable is enabled.
+func getWorkLoop(logger *zap.Logger, criticalityConfig map[string]string, w worker.Worker) (runner, error) {
+	local, err := parseBool(criticalityConfig[configLocal], false)
+	if err != nil {
+		return nil, fmt.Errorf("parsing %q config as bool: %w", configLocal, err)
+	}
+	var workLoop runner
+	if local {
```
maxfisher-g, line 66: nit: the condition can be inverted so that you can unconditionally return and then the `else` branch and `var workLoop` declaration above are not necessary

maxfisher-g (approved):

cmd/collect_signals/main.go
calebbrown: Good point.

cmd/collect_signals/localworker/loop.go
calebbrown: Sure!

cmd/collect_signals/localworker/loop.go
```diff
@@ -0,0 +1,239 @@
...
+
+	if err := l.w.Process(ctx, req, bucketURL); err != nil {
+		return fmt.Errorf("worker process: %w", err)
+	}
+	return nil
+}
+
+// Run iterates through the repositories in batches, calling the supplied
+// worker.Process method for each batch.
+func (l *WorkLoop) Run() error {
```
maxfisher-g, line 150: This function has a lot going on. Would it be worth splitting off some of the logic to a helper function? (e.g. the fast forward/recovery part starting line 162). I'll leave it up to your judgement.

maxfisher-g (dismissed):

cmd/collect_signals/localworker/loop.go
calebbrown: restore logic is in a separate function now.

maxfisher-g (approved):

PR #556, opened 2024-02-26
Update go-graphql-client to fix error handling

calebbrown:
go-graphql-client@v0.11.0 was not wrapping and unwrapping errors correctly.

v0.12.0 fixes this issue.

Furthermore, the logging that would have made this obvious was logging at an "info" level, this change makes the log entry an "error" instead.

### DIFF
--- a/cmd/collect_signals/localworker/loop.go
+++ b/cmd/collect_signals/localworker/loop.go
@@ -215,7 +215,7 @@ func (l *WorkLoop) Run() error {
 				// attempts to the end of the queue to add a delay. Additionally,
 				// if different requests fail consecutively then returning an
 				// error here may be better.
-				logger.With(zap.Error(err), zap.Int("attempt", s.Attempt)).Info("Error processing request")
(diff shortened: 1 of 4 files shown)


cmd/collect_signals/localworker/loop.go
```diff
@@ -215,7 +215,7 @@ func (l *WorkLoop) Run() error {
 				// attempts to the end of the queue to add a delay. Additionally,
 				// if different requests fail consecutively then returning an
 				// error here may be better.
-				logger.With(zap.Error(err), zap.Int("attempt", s.Attempt)).Info("Error processing request")
+				logger.With(zap.Error(err), zap.Int("attempt", s.Attempt)).Error("Error processing request")
```
spencerschrock, line 218: worth updating the comment above?

spencerschrock (approved):

cmd/collect_signals/localworker/loop.go
spencerschrock: never mind, GitHub only showed a few lines of context. I thought the error was referring to the log level

PR #571, opened 2024-03-17
Fix the slightly incorrect docs for running go install.

calebbrown:
After testing, the "/v2" should not be in the package URL during `go install`.

This change fixes the commands in the README.md files.


calebbrown:
Weird. The commands work now.

PR #817, opened 2025-05-07
Fix issue count error

tzing0013:
To get the total number of records by setting per page 1. But the GitHub v3 API is returning ONLY the "next" page (&page=2), so LastPage will be zero since rel="last" is not returned from the API call. That make the issue count zero all the time.

Get issue count with GitHub Search API

### DIFF
--- a/internal/collector/github/legacy/issues.go
+++ b/internal/collector/github/legacy/issues.go
@@ -16,6 +16,7 @@ package legacy
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/google/go-github/v47/github"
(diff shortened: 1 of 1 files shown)


calebbrown:
Thanks for the contribution!

I need to take a look to understand how the switch to the search API changes things.

This change is okay for one-off runs of Criticality Score, but when running it in a batch job we will quickly chew through search query quota, which is separate to REST quota in general.

tzing0013:
> Thanks for the contribution!
> 
> I need to take a look to understand how the switch to the search API changes things.
> 
> This change is okay for one-off runs of Criticality Score, but when running it in a batch job we will quickly chew through search query quota, which is separate to REST quota in general.

I really didn't consider the quota. For REST, the only way I know is the loop in next page, so I change it to search API. Is there any other better way for the count, it's honor to learn it. 

May get the count with GraphQL API,  The lib go-github recommends the library [shurcooL/githubv4](https://github.com/shurcooL/githubv4) for GraphQL.

PR #828, opened 2026-07-10
chore: use NodeID for retrieving user info and use type for detecting bots

### LINKED ISSUE #827: Fails to run this tool because of Copilot account
I tried to use this tool on my repository (`https://github.com/vrc-get/vrc-get`) and failed with `Could not resolve to a User with the login of 'Copilot'. (type: NOT_FOUND), Locations: [], Extensions: map[code:request_error], Path: []"}`.

It looks the `Copilot` account (id: 198982749) does not have `[bot]` suffix on login filed of `/contributors` API endpoint but we cannot use `https://api.github.com/users/Copilot` to access the user though (we should access `https://api.github.com/users/Copilot[bot]`).
It seem a problem of github api so I'm considering report bug to github, but it might be better to use `https://api.github.com/user/$id` access instead of login-based access for reliable access

anatawa12:
Fixes compatibility with copilot account 

Fixes #827 
Closes #824 because it's working for same goal but with different approach


elkampu:
Thanks for tackling this. I tested the response path and found a decoding issue in the current patch: `BatchQuery` uses `ExecRaw` and then ordinary `encoding/json` into `map[string]T`. For `node(id: ...){... on User{company}}`, GitHub returns the fragment fields flat (for example, `{"field0":{"company":"Example Inc."}}`), without a `User` wrapper, so `u.User.Company` remains empty and the organization count becomes 0. A focused HTTP-level test with one human node ID and company reproduces this. A typed `nodes(ids: $ids)` query decoded by `go-graphql-client` handles the inline fragment correctly; it would also be worth filtering for `type == "User"` and skipping empty node IDs. I have a locally validated test/implementation if sharing it would help.

anatawa12:
I could not find any users with company field assigned. Could you let me know some projects with several users with company field assigned please?

elkampu:
Sure. I rechecked current public contributor profiles and found several good examples with multiple non-empty `company` fields:

- `getsentry/sentry-dotnet`: `bruno-garcia` (`@getsentry - sentry.io`), `jamescrosswell` (`Mental Desk Ltd.`), `mattjohnsonpint` (`@cloudflare`), `bitsandfoxes` (`getsentry`), `Flash0ver` (`@FlashOWare`).
- `microsoft/kiota-dotnet`: `baywet` (`@microsoft`), `hwoodiwiss` (`@justeattakeaway`), `MihaMarkic` (`Righthand`), `Ndiritu` (`Microsoft`), `andreaTP` (`IBM`), `svrooij` (`Smartersoft BV`).
- `dotnet/runtime`: `stephentoub`, `jkotas`, `MichalStrehovsky`, and `ericstj` currently list `Microsoft`; `migueldeicaza` lists `Xibbon`.

Those should give the NodeID/user-info path several real profiles to exercise. If you push a follow-up, I’m happy to re-test the HTTP/GraphQL decode path against one of these repositories.

PR #829, opened 2026-07-17
Use the search API to count issues

### LINKED ISSUE #816: There's abnormal for command of 'criticality_score'
I install 'criticality_score' command with the way in [README](https://github.com/ossf/criticality_score?tab=readme-ov-file#usage), and add -depsdev-disable to skip 'gcloud auth login'

I get  the output as follows,  the updated_issues_count, closed_issues_count, issue_comment_frequency is always zero. I think there's some abnormal. Could you help me with this problem.
```shell
$ criticality_score -log=debug -depsdev-disable https://github.com/vllm-project/vllm
2025-04-27 07:13:36.234 INFO    Preparing default scorer
2025-04-27 07:13:36.235 WARN    deps.dev signal source is disabled.
github.com/ossf/criticality_score/v2/internal/collector.New
        /home/openeuler/go/pkg/mod/github.com/ossf/criticality_score/v2@v2.0.4/internal/collector/collector.go:74
main.main
        /home/openeuler/go/pkg/mod/github.com/ossf/criticality_score/v2@v2.0.4/cmd/criticality_score/main.go:164
runtime.main
        /usr/lib/golang/src/runtime/proc.go:267
2025-04-27 07:13:36.235 DEBUG   Parsed project url      {"url": "https://github.com/vllm-project/vllm"}
2025-04-27 07:13:36.235 DEBUG   Fetching basic data from GitHub {"url": "https://github.com/vllm-project/vllm"}
2025-04-27 07:13:36.657 DEBUG   Fetching created time   {"url": "https://github.com/vllm-project/vllm"}
2025-04-27 07:13:36.980 INFO    Collecting      {"url": "https://github.com/vllm-project/vllm", "canonical_url": "https://github.com/vllm-project/vllm"}
2025-04-27 07:13:36.980 DEBUG   Fetching contributors   {"url": "https://github.com/vllm-project/vllm"}
2025-04-27 07:13:37.341 DEBUG   Fetching org count      {"url": "https://github.com/vllm-project/vllm"}
2025-04-27 07:13:38.044 DEBUG   Fetching releases       {"url": "https://github.com/vllm-project/vllm"}
2025-04-27 07:13:38.457 DEBUG   Fetching closed issues  {"url": "https://github.com/vllm-project/vllm"}
2025-04-27 07:13:38.844 DEBUG   Fetching updated issues {"url": "https://github.com/vllm-project/vllm"}
repo.url: https://github.com/vllm-project/vllm
repo.language: Python
repo.license: Apache License 2.0
repo.star_count: 45865
repo.created_at: 2023-02-09T11:23:20Z
repo.updated_at: 2025-04-27T05:53:23Z
legacy.created_since: 26
legacy.updated_since: 0
legacy.contributor_count: 1022
legacy.org_count: 8
legacy.commit_frequency: 94.08
legacy.recent_release_count: 34
legacy.updated_issues_count: 0
legacy.closed_issues_count: 0
legacy.issue_comment_frequency: 0
legacy.github_mention_count: 1606
default_score: 0.62051
```

camUrban:
## Summary

Fixes #816.

The three legacy issue signals (`updated_issues_count`, `closed_issues_count`, and `issue_comment_frequency`) currently report 0 for every repository.

GitHub's repo issues list endpoint (`/repos/{owner}/{repo}/issues`) now uses cursor-based pagination: its `Link` header contains only a `rel="next"` entry with an opaque cursor, and no `rel="last"` entry. `FetchIssueCount` relied on requesting `per_page=1` and reading the total from the last page number, so it now always sees `LastPage == 0` and returns 0. Since `issue_comment_frequency` divides the comment count by the updated-issues count, it is zeroed as well.

This change counts issues with a single search API request instead, reading the exact `total_count` from the response. The original semantics are preserved:

- Both issues and pull requests are counted (search includes both by default, matching the list endpoint).
- Results are filtered on updated time within the lookback window (`updated:>=<timestamp>`), matching the list endpoint's `since` parameter.
- The closed state is expressed as `is:closed`, matching `state=closed` combined with `since` (closed-state items updated within the window).
- The 5xx fallback to `MaxIssuesLimit` is unchanged.

`FetchIssueCommentCount` is untouched. The comments endpoint still uses page-number pagination.

## Testing

- `make test` passes.
- `make lint` passes.
- Verified end to end against `camUrban/PteraSoftware` (a repo with recent issue activity): before this change the three signals were 0/0/0; after, they are 117/100/1.19, matching direct queries to the search API for the same 90-day window.

### DIFF
--- a/internal/collector/github/legacy/issues.go
+++ b/internal/collector/github/legacy/issues.go
@@ -16,6 +16,7 @@ package legacy
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/google/go-github/v47/github"
@@ -36,12 +37,18 @@ const (
 //
 // This count includes both issues and pull requests.
 func FetchIssueCount(ctx context.Context, c *githubapi.Client, owner, name string, state IssueState, lookback time.Duration) (int, error) {
-	opts := &github.IssueListByRepoOptions{
-		Since:       time.Now().UTC().Add(-lookback),
-		State:       string(state),
-		ListOptions: github.ListOptions{PerPage: 1}, // 1 result per page means LastPage is total number of records.
+	// The repo issues list endpoint uses cursor-based pagination, which does
+	// not expose the last page number, so the count comes from the search
+	// API's total_count instead.
+	since := time.Now().UTC().Add(-lookback)
+	query := fmt.Sprintf("repo:%s/%s updated:>=%s", owner, name, since.Format(time.RFC3339))
+	if state != IssueStateAll {
+		query = fmt.Sprintf("%s is:%s", query, state)
 	}
-	is, resp, err := c.Rest().Issues.ListByRepo(ctx, owner, name, opts)
+	opts := &github.SearchOptions{
+		ListOptions: github.ListOptions{PerPage: 1},
+	}
+	res, _, err := c.Rest().Search.Issues(ctx, query, opts)
 	// The API returns 5xx responses if there are too many issues.
 	if c := githubapi.ErrorResponseStatusCode(err); 500 <= c && c < 600 {
 		return MaxIssuesLimit, nil
@@ -49,10 +56,7 @@ func FetchIssueCount(ctx context.Context, c *githubapi.Client, owner, name strin
 	if err != nil {
 		return 0, err
 	}
-	if resp.NextPage == 0 {
-		return len(is), nil
-	}
-	return resp.LastPage, nil
+	return res.GetTotal(), nil
 }
 
 // FetchIssueCommentCount returns the total number of comments for a given repo


ruohai0925:
+1 on this — I hit the same bug independently and can confirm both the diagnosis
and the impact.

The `Link` header on `/repos/{owner}/{repo}/issues` now looks like this (note
the opaque `after=` cursor and the complete absence of `rel="last"`):

```
Link: <https://api.github.com/repositories/150626842/issues?state=all&since=2026-05-04T00%3A00%3A00Z&per_page=1&after=Y3Vyc29yOnYyOpLPAAABn8D-mnjPAAAAASx2iuA%3D&page=2>; rel="next"
```

so `resp.LastPage` is `0` and `FetchIssueCount` returns `0` unconditionally.

Some measurements on top of the `camUrban/PteraSoftware` numbers already in the
description, in case they are useful for review — `criticality_score
-depsdev-disable`, before vs. after fixing the count:

| repo | `updated_issues` | `closed_issues` | `comment_frequency` | `default_score` |
|---|---:|---:|---:|---:|
| `BLAST-WarpX/warpx` | 0 → 690 | 0 → 572 | 0 → 0.5 | 0.53654 → **0.62253** |
| `ruohai0925/IAMReX` | 0 → 2 | 0 → 2 | 0 → 0 | 0.38916 → **0.40145** |

The 690/572 figures match what the search API reports for the same 90-day
window, which is a nice independent check on the approach taken here.

Worth emphasising how much this distorts things: WarpX moves by **+0.086**. The
three affected signals carry 2.0 of the 10.5 total weight in
`original_pike.yml`, so every published score for an active project is currently
biased low, and the more issue activity a project has the more it loses.

Two small things I noticed while looking at the same code, both arguing that the
scope here is right:

- `FetchIssueCommentCount` (`/issues/comments`) and `FetchTotalContributors`
  (`/contributors`) do still return `rel="last"` (I see `page=347` and `page=65`
  respectively), so leaving them on the `per_page=1` trick is correct for now.
- `internal/collector/github/legacy` has no test coverage, so there is nothing
  guarding against this class of regression. I wrote `httptest`-backed tests for
  `FetchIssueCount` while investigating (cursor fallback, `rel="last"` fast
  path, 5xx handling, the `MaxIssuesLimit` cap, the `since` window) — happy to
  send them as a follow-up to this PR if that would help it land.

I also opened #830 with a paging-based fix before seeing how far along this one
was. The search API approach here is cleaner and I would rather see this one
merged — I'll close mine if this moves.

